PR #11, opened 2022-07-18
Optional: cli tool to help authors calculate method numbers

alexytsu:
Not really necessary for this util to live in this repo, but I created a package to quickly calculate method numbers from the CLI

This will be helpful later when testing/verifying the ouput of the fvm_dispatch methods/macros in integration tests or on-chain


Cargo.toml
```diff
@@ -2,5 +2,6 @@
 
 members = [
     "fvm_dispatch",
+    "fvm_dispatch_helper",
```
anorth, line 5: `fvm_dispatch_tools`? Helper is rather generic. There's quite likely more functionality we can add to tools for use at build time, such as detecting collisions.

fvm_dispatch_helper/src/blake2b.rs
```diff
@@ -0,0 +1,11 @@
+use fvm_dispatch::hash::Hasher;
+use multihash::Code;
+use multihash::MultihashDigest;
+pub struct Blake2bHasher {}
+impl Hasher for Blake2bHasher {
+    fn hash(&self, bytes: &[u8]) -> Vec<u8> {
+        let digest = Code::Blake2b256.digest(bytes);
+        // drop the first 4 bytes of the multihash which identify the hash as blake2b256
```
anorth, line 8: The multihash library is probably too high level for this dependency. Find out what it depends on to implement blake2b and use that instead.

anorth (commented):
Good start. I suggest making this a bit easier to embed in other tooling. Some thoughts:
- Make default output "clean" of labels etc for easier scripting
- Read names on stdin lines, print method numbers on stdout (but a CLI arg for a one-shot is good too)

anorth (approved):

PR #20, opened 2022-07-19
Refactor and rename fil_token state layer

alexytsu:
Following up with a couple of housekeeping items from the last PR

- rename fil_token -> fil_fungible_token
- fungible_token: rename some public structs
- panic in load_state if state tree missing
- use a shared-memory blockstore for testing
- streamline balance and allowance increase/decreases into single 'change_by' functions


fil_fungible_token/src/blockstore.rs
```diff
@@ -42,31 +44,51 @@ impl fvm_ipld_blockstore::Blockstore for Blockstore {
     }
 }
 
-// TODO: put this somewhere more appropriate when a tests folder exists
-/// An in-memory blockstore impl taken from filecoin-project/ref-fvm
+/// An in-memory blockstore impl that shares underlying memory when cloned
```
alexytsu, line 47: I _think_ it is ok to assume this is how a blockstore implementation should work. a blockstore impl running on-chain should persist everything to the ipld layer directly making the actual blockstore handle "stateless" from the pov of a clone

fil_fungible_token/src/token/mod.rs
```diff
@@ -13,27 +15,31 @@ use fvm_ipld_blockstore::Blockstore as IpldStore;
 use fvm_shared::bigint::Zero;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::ActorID;
+use num_traits::Signed;
 
 /// Library functions that implement core FRC-??? standards
 ///
 /// Holds injectable services to access/interface with IPLD/FVM layer.
-pub struct TokenHelper<BS>
+pub struct Token<BS>
```
alexytsu, line 23: Naming wise went with calling the top-level library that authors will use `Token` 

and the trait/interface that they will implement is defined in `token/types.rs` as `FrcXXXToken`

fil_fungible_token/src/token/state.rs
```diff
@@ -494,33 +373,104 @@ mod test {
...
+        let actor: ActorID = 1;
+
+        // can't decrease from zero
+        let err = state
+            .change_balance_by(bs, actor, &BigInt::from(-1))
+            .unwrap_err();
+        assert_eq!(
+            err.to_string(),
+            "resulting balance was negative adding -1 to 0"
+        );
```
alexytsu, line 416: testing against the error msg here seems a little imprecise, probably indicating the misuse or innappropriateness of `anyhow` within this library.

abright: Probably better off using `thiserror` to define more concrete error types we can match against. In this case especially it could mean having a nice "insufficient balance" error or something which included the values.

Both crates suggest using `thiserror` for library-style code:
https://github.com/dtolnay/thiserror#comparison-to-anyhow
https://github.com/dtolnay/anyhow#comparison-to-thiserror

fil_fungible_token/src/token/mod.rs
```diff
@@ -43,22 +48,25 @@ where
...
+    ///
+    /// Actors can't usefully recover if state wasn't initialized (failure to call `init_state`) in
+    /// the constructor so this method panics if the state tree if missing
+    fn load_state(&self) -> TokenState {
+        TokenState::load(&self.bs, &self.state_cid).unwrap()
     }
 
     /// Mints the specified value of tokens into an account
     ///
     /// If the total supply or account balance overflows, this method returns an error. The mint
```
anorth, line 60: Nit: remove reference to overflow

fil_fungible_token/src/token/state.rs
```diff
@@ -62,17 +63,14 @@ impl TokenState {
...
     }
 
     /// Saves the current state to the blockstore, returning the cid
-    /// TODO: Potentially should replaced with more targeted saving of different branches of the state tree for
-    /// efficiency
-    /// - possibly keep track of "dirty" branches that need to be flushed and put into the blockstore
-    /// on save
-    /// - hack for now is that each method that touches a hamt, will itself commit the changes to the blockstore
+    /// TODO: should replaced with more targeted saving of different branches of the state tree for efficiency
+    /// i.e. only save the balances HAMT if it has changed, only save the allowance HAMTs if they have changed
```
anorth, line 73: I *think* this is going to do the right thing anyway. But maybe write an issue rather than TODO here

anorth (approved):
This structure is looking really good. At least until we try to do receiver hooks :-).

Still more test cases to come, I think, e.g. around the cleaning up of empty entries. Happy for error improvements to come in a follow-up.

Issue #15, opened 2022-07-19
Evaluate dropping the dependency on anyhow


alexytsu:
Discussions of dropping anyhow in some of our dependencies may warrant re-evaluating if we want to opt our consumers in to this
https://github.com/filecoin-project/builtin-actors/pull/214
https://github.com/filecoin-project/ref-fvm/pull/463

alexytsu:
Closed by https://github.com/helix-onchain/filecoin/pull/22

PR #22, opened 2022-07-20
custom error enums for state and token methods

alexytsu:
Replaces use internal use of anyhow with custom error structs

The `fvm_ipld_blockstore::Blockstore` trait depends exports an interface with `anyhow::Result` so it can't be dropped as a dependency for now.


alexytsu:
Closes https://github.com/helix-onchain/filecoin/issues/15

fil_fungible_token/src/token/state.rs
```diff
@@ -1,25 +1,52 @@
...
+#[derive(Error, Debug)]
+pub enum StateError {
+    #[error("ipld hamt error: {0}")]
+    IpldHamt(#[from] HamtError),
+    #[error("missing state at cid: {0}")]
+    MissingState(Cid),
+    #[error("underlying serialization error: {0}")]
+    Serialization(String),
+    #[error(
+        "negative balance caused by subtracting {delta:?} from {owner:?}'s balance of {balance:?}"
```
anorth, line 30: The `delta` here will be negative when subtracting, so this double negative will be confusing.

Instead: "negative balance caused by _adding_ -1000 to owner's balance of 200". Or "changing balance by xxx"

anorth (approved):
Great that you got a net negative lines of code with this, too.

PR #24, opened 2022-07-20
Token tests

alexytsu:
- unit tests


alexytsu:
CleanShot 2022-07-20 at 18 16 03

Side note: worth adding code coverage reports to CI?

fil_fungible_token/src/token/mod.rs
```diff
@@ -40,18 +40,16 @@ impl<BS> Token<BS>
...
+        Self { bs, state_cid }
     }
 
     /// Constructs the token state tree and saves it at a CID
-    pub fn init_state(&self) -> Result<Cid> {
+    pub fn init_state(&mut self) -> Result<Cid> {
         let init_state = TokenState::new(&self.bs)?;
-        Ok(init_state.save(&self.bs)?)
+        self.state_cid = init_state.save(&self.bs)?;
+        Ok(self.state_cid)
```
anorth, line 52: Hooray for tests finding bugs!

Use `self.save_state()` ?

fil_fungible_token/src/token/mod.rs
```diff
@@ -215,10 +219,13 @@ where
             state.attempt_use_allowance(&self.bs, spender, owner, &value)?;
         }
         // attempt to burn the requested amount
-        let new_amount = state.change_balance_by(&self.bs, owner, &value.neg())?;
+        let new_amount = state.change_balance_by(&self.bs, owner, &value.clone().neg())?;
+
+        // decrease total_supply
+        state.change_supply_by(&value.neg())?;
```
anorth, line 225: Nit: extract `value.neg()` to compute negation only once. I'm also curious why a clone was required on 222.

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+    use num_traits::Zero;
+
+    use crate::blockstore::SharedMemoryBlockstore;
+
+    use super::Token;
+
+    #[test]
+    fn it_instantiates() {
+        let state_cid = Cid::default();
+        let mut token = Token::new(SharedMemoryBlockstore::new(), state_cid);
```
anorth, line 306: This is weird constructing with an invalid state CID. I think maybe we can find better construction patterns. Instead of `new`, I suggest adding `create()`, which creates a new state and returns a `Token` referencing it, and then only using `new` with a valid CID (but it probably won't actually check).

A bigger change would hold the `TokenState` member in the `Token`, and then we would rename `new` to `load` and it would actually enforce correct internal state. I'm unsure whether that's worthwhile tho.

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+    use fvm_shared::econ::TokenAmount;
+    use num_traits::Zero;
+
+    use crate::blockstore::SharedMemoryBlockstore;
+
+    use super::Token;
+
+    #[test]
+    fn it_instantiates() {
+        let state_cid = Cid::default();
```
anorth, line 305: Inline this

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+        token.init_state().expect("state init failed");
+        token.load_state();
+    }
+
+    #[test]
+    fn it_persists_loads() {
+        let state_cid = Cid::default();
+        let bs = SharedMemoryBlockstore::new();
+        let mut token = Token::new(bs.clone(), state_cid);
+        token.init_state().unwrap();
```
anorth, line 316: I don't think this line is testing anything new (and also wierd construction as mentioned above)

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+
+        // simulate actor calls mint method
+        let treasury = 1;
+        token.mint(treasury, TokenAmount::from(100)).unwrap();
+        let balance = token.balance_of(treasury).unwrap();
+        assert_eq!(balance, TokenAmount::from(100));
+
+        // here the actor would save the cid to it's state tree
+        // TODO: state_cid needs to be exported so that actors can keep track of it
+        let new_state_cid = token.state_cid;
```
anorth, line 326: A pattern I'd expect here is `new_state_cid = token.save()` (even if, at the moment, save is no-op because it already saved)

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+        token.init_state().unwrap();
+
+        let treasury = 1;
+        token.mint(treasury, TokenAmount::from(1_000_000)).unwrap();
+
+        let balance = token.balance_of(treasury).unwrap();
+        assert_eq!(balance, TokenAmount::from(1_000_000));
+
+        let total_supply = token.total_supply();
+        assert_eq!(total_supply, TokenAmount::from(1_000_000));
```
anorth, line 348: - Mint again to (a) the same address, and (b) a different address and check balances and supply
- Mint zero
- Attempt to mint negative

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+    fn it_burns() {
+        let state_cid = Cid::default();
+        let mut token = Token::new(SharedMemoryBlockstore::new(), state_cid);
+        token.init_state().unwrap();
+
+        let treasury = 1;
+        let mint_amount = TokenAmount::from(1_000_000);
+        let burn_amount = TokenAmount::from(600_000);
+        token.mint(treasury, mint_amount).unwrap();
+        token.burn(treasury, treasury, burn_amount).unwrap();
```
anorth, line 361: - Add a second account and check it's balance doesn't change
- Burn zero
- Burn exact balance
- Attempt to burn negative

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+        let total_supply = token.total_supply();
+        assert_eq!(total_supply, TokenAmount::from(400_000));
+
+        // treasury balance decreased
+        let balance = token.balance_of(treasury).unwrap();
+        assert_eq!(balance, TokenAmount::from(400_000));
+    }
+
+    #[test]
+    fn it_allows_delegated_burns() {
```
anorth, line 373: Nit: move complicated tests like delegation below simple ones like negatives and other edge cases.

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+        assert_eq!(new_allowance, TokenAmount::zero());
+    }
+
+    #[test]
+    fn it_cannot_burn_below_zero() {
+        let state_cid = Cid::default();
+        let mut token = Token::new(SharedMemoryBlockstore::new(), state_cid);
+        token.init_state().unwrap();
+
+        let treasury = 1;
```
anorth, line 424: Consider some constants for addresses to move these lines out of the tests themselves.

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+            .transfer(owner, owner, receiver, TokenAmount::from(60))
+            .unwrap();
+
+        // owner has 100 - 60 = 40
+        let balance = token.balance_of(owner).unwrap();
+        assert_eq!(balance, TokenAmount::from(40));
+
+        // receiver has 0 + 60 = 60
+        let balance = token.balance_of(receiver).unwrap();
+        assert_eq!(balance, TokenAmount::from(60));
```
anorth, line 461: - Test transfer of zero
- Attempt transfer negative
- Transfer to self with zero, and non-zero amounts
- Check the total supply is unchanged in all these cases

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+        let owner = 1;
+        let receiver = 2;
+        let spender = 3;
+
+        // mint 100 for the owner
+        token.mint(owner, TokenAmount::from(100)).unwrap();
+        // approve 100 spending allowance for spender
+        token
+            .increase_allowance(owner, spender, TokenAmount::from(100))
+            .unwrap();
```
anorth, line 556: Aside: I hate Rust's enforced line break heuristics. We can consider "fixing" it like https://github.com/filecoin-project/builtin-actors/blob/master/rustfmt.toml

fil_fungible_token/src/token/mod.rs
```diff
@@ -278,8 +285,380 @@ where
...
+        assert_eq!(receiver_balance, TokenAmount::zero());
+        assert_eq!(spender_balance, TokenAmount::zero());
+
+        // verify allowance was not spent
+        let spender_allowance = token.allowance(owner, spender).unwrap();
+        assert_eq!(spender_allowance, TokenAmount::from(40));
+    }
+
+    #[test]
+    fn it_checks_for_invalid_negatives() {
```
anorth, line 637: Ah, I see. I think better to roll these in with the other tests for each functionality.

anorth (approved):
These are great. I made comments requesting a bunch more tests. I'm happy for you to merge this and follow-up, or make changes here and re-request review.

fil_fungible_token/src/token/mod.rs
alexytsu: That looks good, I'll change the rules in a follow-up to minimise noise on this diff

Issue #16, opened 2022-07-19
Batching or transaction-like capability for exported library methods


alexytsu:
@anorth has suggested

> One consideration for improving the API is efficiency of state writes for bulk operations. The current API will force some unnecessary state.save() calls for intermediate states. E.g. intialising and then minting to a bunch of accounts up front, or making multiple transfers (from some non-standard token method). This API is pretty good for the basic cases and we want to keep them easy. We should think about how/whether to change this API to aid batching, or add an alternative style for less common but more complex calls. Maybe a closure that executes over state object, or a builder-style API that can abstract over it while batching.

alexytsu:
https://github.com/helix-onchain/filecoin/pull/40 is in progress but deferred for now

PR #25, opened 2022-07-21
refactor token state management

alexytsu:
Before addressing the cleanup of tests and additional test-cases from the feedback in https://github.com/helix-onchain/filecoin/pull/24, I wanted to refactor the API provided by `Token` for loading and saving itself to the blockstore.

There was [some awkwardness noted](https://github.com/helix-onchain/filecoin/pull/24#discussion_r926077476) around the `Token::new` pattern. This change addresses that with the following options:
- New tokens are created with `Token::new`
- Existing tokens are loaded from state with `Token::load`

Previously, every call to a method on Token would cause state to be loaded and written to the blockstore. Now, by holding a TokenState rather than a Cid pointing to it, we can cache state changes in memory. This pushes the onus of persistance onto the library consumer, requiring a call to `Token::flush` to persist state changes into the blockstore. This should be a more flexible API allowing greater control of when expensive read/writes occur and enables batching of transactions. 

This incidentally highlights a previously uncaught bug in `transfer` and `burn` which made multiple sequential calls to change `TokenState`. If any step within the transaction fails, previous changes would need to be reverted. This is much easier done if changes are made in memory rather than being committed to the blockstore. To help this, a `Token::transaction` method is introduced to help us write atomic transactions.

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -42,16 +42,15 @@ where
 {
     /// Creates a new token instance using the given blockstore and creates a new empty state tree
     ///
-    /// Returns a Token handle that can be used to interact with the token state tree and the Cid
-    /// of the state tree root
-    pub fn new(bs: BS) -> Result<(Self, Cid)> {
+    /// Returns a Token handle that can be used to interact with the token state tree. The state
+    /// tree is not persisted to the blockstore until `flush` is called.
+    pub fn new(bs: BS) -> Result<Self> {
         let init_state = TokenState::new(&bs)?;
-        let cid = init_state.save(&bs)?;
         let token = Self {
             bs,
             state: init_state,
         };
-        Ok((token, cid))
+        Ok(token)
     }
 
     /// For an already initialised state tree, loads the state tree from the blockstore and returns
@@ -77,7 +76,7 @@ where
     {
         let mut mutable_state = self.state.clone();
         let res = f(&mut mutable_state, self.bs.clone())?;
-        // if closure didn't error, save state
+        // if closure didn't error, save to state cache which will be persisted to blockstore on the next call of flush
         self.state = mutable_state;
         Ok(res)
     }
@@ -99,12 +98,16 @@ where
         }
 
         // Increase the balance of the actor and increase total supply
-        self.state
-            .change_balance_by(&self.bs, initial_holder, &value)?;
+        self.transaction(|state, bs| {
+            state.change_balance_by(&bs, initial_holder, &value)?;
 
-        // TODO: invoke the receiver hook on the initial_holder
+            // TODO: invoke the receiver hook on the initial_holder
 
-        self.state.change_supply_by(&value)?;
+            state.change_supply_by(&value)?;
+
+            Ok(())
+        })?;
+
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


alexytsu:
https://github.com/helix-onchain/filecoin/issues/16

By putting the responsibility of calling `flush` on the consumers, they can implement their own batching methods

fil_fungible_token/src/token/mod.rs
```diff
@@ -32,39 +32,61 @@ where
...
-    /// Instantiate an empty token helper with access to a blockstore and runtime
-    pub fn new(bs: BS, state_cid: Cid) -> Self {
-        Self { bs, state_cid }
+    /// Creates a new token instance using the given blockstore and creates a new empty state tree
+    ///
+    /// Returns a Token handle that can be used to interact with the token state tree and the Cid
+    /// of the state tree root
+    pub fn new(bs: BS) -> Result<(Self, Cid)> {
+        let init_state = TokenState::new(&bs)?;
+        let cid = init_state.save(&bs)?;
```
anorth, line 49: You discarded the Cid return value at every usage. An alternative here is to not save the state here and require a `flush` (after any subsequent modifications).

fil_fungible_token/src/token/mod.rs
```diff
@@ -32,39 +32,61 @@ where
...
+    ///
+    /// If the closure returns an error, the transaction is dropped atomically and no change is
+    /// observed on token state.
+    pub fn transaction<F, Res>(&mut self, f: F) -> Result<Res>
+    where
+        F: FnOnce(&mut TokenState, BS) -> Result<Res>,
+    {
+        let mut mutable_state = self.state.clone();
+        let res = f(&mut mutable_state, self.bs.clone())?;
+        // if closure didn't error, save state
```
anorth, line 80: "save" is misleading here since you don't flush. Clarify in the comment that the state isn't flushed at the end of the transaction.

fil_fungible_token/src/token/mod.rs
```diff
@@ -32,39 +32,61 @@ where
...
-    /// Helper function that saves the state tree and returns the CID of the root
-    fn save_state(&mut self, state: &TokenState) -> Result<Cid> {
-        self.state_cid = state.save(&self.bs)?;
-        Ok(self.state_cid)
+    /// Opens an atomic transaction on TokenState which allows a closure to make multiple
+    /// modifications to the state tree.
+    ///
+    /// If the closure returns an error, the transaction is dropped atomically and no change is
+    /// observed on token state.
+    pub fn transaction<F, Res>(&mut self, f: F) -> Result<Res>
```
anorth, line 74: This is interesting but not quite perfect.
1. Ideally don't expose the token state directly to Token users
2. Users can still use the API outside a transaction.

I think this method makes a pretty good _internal_ transaction helper. I.e. call it from within Token methods so they are atomic, but don't expose it publicly. So I reckon make it non-`pub` for this PR, fixing (1).

I'm not sure 2 is a serious problem, but would be _nice_ to address. For application-level transactions, we'd need the callback parameter to be Token or a new TokenTxn trait. If callers strictly used that, it would kinda give (2).

fil_fungible_token/src/token/mod.rs
```diff
@@ -77,15 +99,12 @@ where
...
+        self.state
+            .change_balance_by(&self.bs, initial_holder, &value)?;
 
         // TODO: invoke the receiver hook on the initial_holder
 
-        state.change_supply_by(&value)?;
-
-        // Commit the state atomically if supply and balance increased
-        self.save_state(&state)?;
+        self.state.change_supply_by(&value)?;
```
anorth, line 107: Use a txn for these two mutations

fil_fungible_token/src/token/mod.rs
```diff
@@ -266,76 +279,99 @@ where
             )));
         }
 
-        let mut state = self.load_state();
+        let _old_state = self.state.clone();
```
anorth, line 282: Unused (for now)

anorth (approved):

PR #28, opened 2022-07-22
Call receiver hooks

alexytsu:
Final bit of functionality before returning to tidy up the unit tests

Calls receiver hooks during minting and transfer, reverting the operation if the hook aborts

### DIFF
--- a/fil_fungible_token/src/method.rs
+++ b/fil_fungible_token/src/method.rs
@@ -1,6 +1,8 @@
 use fvm_ipld_encoding::Error as IpldError;
 use fvm_ipld_encoding::RawBytes;
 use fvm_sdk::{send, sys::ErrorNumber};
+use fvm_shared::error::ExitCode;
+use fvm_shared::receipt::Receipt;
 use fvm_shared::{address::Address, econ::TokenAmount, ActorID};
 use num_traits::Zero;
 use thiserror::Error;
@@ -19,31 +21,30 @@ pub enum MethodCallError {
 
 /// An abstraction used to send messages to other actors
 pub trait MethodCaller {
-    /// Call the receiver hook on a given actor
+    /// Call the receiver hook on a given actor, specifying the amount of tokens pending to be sent
+    /// and the sender and receiver
     ///
     /// Returns true if the receiver hook is called and exits without error, else returns false
     fn call_receiver_hook(
         &self,
         from: ActorID,
         to: ActorID,
-        value: &TokenAmount,
+        token_value: &TokenAmount,
         data: &[u8],
-    ) -> Result<bool>;
+    ) -> Result<Receipt>;
 }
 
 #[derive(Debug, Default, Clone, Copy)]
 pub struct FvmMethodCaller {}
 
-// TODO: this should be moved to the example actor so the dependency on fvm_sdk is dropped from the
-// library
 impl MethodCaller for FvmMethodCaller {
     fn call_receiver_hook(
         &self,
         from: ActorID,
         to: ActorID,
-        value: &TokenAmount,
+        token_value: &TokenAmount,
         data: &[u8],
-    ) -> Result<bool> {
+    ) -> Result<Receipt> {
         // TODO: use fvm_dispatch here (when it supports compile time method resolution)
         // TODO: ^^ necessitates determining conventional method names for receiver hooks
 
@@ -54,14 +55,12 @@ impl MethodCaller for FvmMethodCaller {
 
         let params = TokenReceivedParams {
             sender: Address::new_id(from),
-            value: value.clone(),
+            value: token_value.clone(),
             data: RawBytes::from(data.to_vec()),
         };
         let params = RawBytes::new(fvm_ipld_encoding::to_vec(&params)?);
 
-        let receipt = send::send(&to, METHOD_NUM, params, TokenAmount::zero())?;
-
-        Ok(receipt.exit_code.is_success())
+        Ok(send::send(&to, METHOD_NUM, params, TokenAmount::zero())?)
     }
 }
 
@@ -79,11 +78,19 @@ impl MethodCaller for FakeMethodCaller {
         _to: ActorID,
         _value: &TokenAmount,
         data: &[u8],
-    ) -> Result<bool> {
+    ) -> Result<Receipt> {
         if data.is_empty() {
-            Ok(true)
+            Ok(Receipt {
+                exit_code: ExitCode::OK,
+                return_data: RawBytes::new(data.to_vec()),
+                gas_used: 0,
+            })
         } else {
-            Ok(false)
+            Ok(Receipt {
+                exit_code: ExitCode::SYS_INVALID_RECEIVER,
+                return_data: RawBytes::new(data.to_vec()),
+                gas_used: 0,
+            })
         }
     }
 }
--- a/fil_fungible_token/src/method.rs
+++ b/fil_fungible_token/src/method.rs
@@ -80,7 +80,7 @@ impl MethodCaller for FakeMethodCaller {
         _value: &TokenAmount,
         data: &[u8],
     ) -> Result<bool> {
-        if data.len() == 0 {
+        if data.is_empty() {
             Ok(true)
         } else {
             Ok(false)
--- a/fil_fungible_token/src/method.rs
+++ b/fil_fungible_token/src/method.rs
@@ -1,0 +1,89 @@
+use fvm_ipld_encoding::Error as IpldError;
+use fvm_ipld_encoding::RawBytes;
+use fvm_sdk::{send, sys::ErrorNumber};
+use fvm_shared::{address::Address, econ::TokenAmount, ActorID};
+use num_traits::Zero;
+use thiserror::Error;
+
+use crate::receiver::types::TokenReceivedParams;
+
+type Result<T> = std::result::Result<T, MethodCallError>;
+
+#[derive(Error, Debug)]
+pub enum MethodCallError {
+    #[error("fvm syscall error: `{0}`")]
+    Syscall(#[from] ErrorNumber),
+    #[error("ipld serialization error: `{0}`")]
+    Ipld(#[from] IpldError),
+}
+
+/// An abstraction used to send messages to other actors
+pub trait MethodCaller {
+    /// Call the receiver hook on a given actor
+    ///
+    /// Returns true if the receiver hook is called and exits without error, else returns false
+    fn call_receiver_hook(
+        &self,
+        from: ActorID,
+        to: ActorID,
+        value: &TokenAmount,
+        data: &[u8],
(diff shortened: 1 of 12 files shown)


fil_fungible_token/src/token/types.rs
```diff
@@ -1,5 +1,4 @@
 use anyhow::Result;
```
alexytsu, line 1: This should be replaced with custom errors in a followup

fil_fungible_token/src/token/types.rs
```diff
@@ -1,12 +1,13 @@
...
```
alexytsu, line 1: remove this use of anyhow in a followup

fil_fungible_token/src/receiver/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod types;
```
alexytsu, line 1: TODO: library level code for the receiving actor (don't think there'll be _much_ boilerplate tho)

maybe receiver module should be it's own package?

fil_fungible_token/src/token/mod.rs
```diff
@@ -272,71 +318,92 @@ where
             )));
         }
 
-        let _old_state = self.state.clone();
+        let old_state = self.state.clone();
```
alexytsu, line 321: thoughts on alternative patterns for "reverting" here? 

the state struct itself should be rather light however, holding Cids, not hamts

fil_fungible_token/src/method.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug, Default, Clone, Copy)]
+pub struct FvmMethodCaller {}
+
+// TODO: this should be moved to the example actor so the dependency on fvm_sdk is dropped from the
+// library
+impl MethodCaller for FvmMethodCaller {
+    fn call_receiver_hook(
+        &self,
+        from: ActorID,
+        to: ActorID,
```
anorth, line 43: This exposes an interesting chicken and egg problem.

In the proposal, I wrote:
> An uninitialised pubkey address can receive tokens (because receiver hook will initialise it), but cannot be authorized as a spender

I was about to say you won't always have the actor ID so change to address, but I realise that we can't have the token state reflecting the new balances persisted if we don't know the receiving actor ID.

I think the required workaround will be that address resolution (in the actor code) needs to initialise actors if necessary. See `resolve_to_id_addr` in the builtin-actors `runtime/src/builtin/shared.rs`. And then this API is ok. (Defer this for a follow-up)

fil_fungible_token/src/method.rs
```diff
@@ -0,0 +1,89 @@
...
+        value: &TokenAmount,
```
anorth, line 44: This initially confused me (I thought it was going to transfer native FIL tokens). Add documentation.

fil_fungible_token/src/method.rs
```diff
@@ -0,0 +1,89 @@
...
+        let to = Address::new_id(to);
+
+        let params = TokenReceivedParams {
+            sender: Address::new_id(from),
+            value: value.clone(),
+            data: RawBytes::from(data.to_vec()),
+        };
+        let params = RawBytes::new(fvm_ipld_encoding::to_vec(&params)?);
+
+        let receipt = send::send(&to, METHOD_NUM, params, TokenAmount::zero())?;
```
anorth, line 62: There's a policy decision to be made about what to do with `NotFound` syscall error. Replace the `?` with explicit error handling in this case (even if for now we treat all errors as failure).

Further, I think we want this logic encapsulated in the token logic, both for testing and reliable implementation. So I reckon change the `call_receiver_hook` return value to expose the return value from `send::send` so there's no logic ending up here.

fil_fungible_token/src/method.rs
```diff
@@ -0,0 +1,89 @@
...
+        Ok(receipt.exit_code.is_success())
+    }
+}
+
+/// A fake method caller that can simulate the receiving actor return true or false
+///
+/// If call_receiver_hook is called with an empty data array, it will return true.
+/// If call_receiver_hook is called with a non-empty data array, it will return false.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct FakeMethodCaller {}
```
anorth, line 73: Nit: we'll need a mock that can make assertions about expected values for all the parameters, and have dynamic behaviour in response, for subsequent testing.

fil_fungible_token/src/token/mod.rs
```diff
@@ -19,43 +18,53 @@ pub enum TokenError {
     State(#[from] StateError),
     #[error("invalid negative: {0}")]
     InvalidNegative(String),
+    #[error("error calling receiver hook: {0}")]
+    MethodCall(#[from] MethodCallError),
+    #[error("receiver hook aborted when {from:?} sent {value:?} to {to:?} by {by:?}")]
+    ReceiverHook { from: ActorID, to: ActorID, by: ActorID, value: TokenAmount },
```
anorth, line 24: Can we capture the exit code it aborted with too?

fil_fungible_token/src/token/mod.rs
```diff
@@ -19,43 +18,53 @@ pub enum TokenError {
...
+    MC: MethodCaller,
 {
     /// Injected blockstore. The blockstore must reference the same underlying storage under Clone
     bs: BS,
+    /// Minimal interface to call methods on other actors (i.e. receiver hooks)
+    mc: MC,
     /// In-memory cache of the state tree
     state: TokenState,
+    /// Token-contract address
+    actor_id: ActorID,
```
anorth, line 44: Consider encapsulating this in `MC`.

fil_fungible_token/src/token/mod.rs
```diff
@@ -19,43 +18,53 @@ pub enum TokenError {
...
+    ReceiverHook { from: ActorID, to: ActorID, by: ActorID, value: TokenAmount },
 }
 
 type Result<T> = std::result::Result<T, TokenError>;
 
 /// Library functions that implement core FRC-??? standards
 ///
 /// Holds injectable services to access/interface with IPLD/FVM layer.
-pub struct Token<BS>
+pub struct Token<BS, MC>
```
anorth, line 32: Note: I don't know if this is a good suggestion, but an alternative here is to pull together all the external needs under a single trait, and then a single trait parameter here.

fil_fungible_token/src/token/mod.rs
```diff
@@ -80,28 +89,64 @@ where
...
+        data: &[u8],
+    ) -> Result<()> {
         if value.is_negative() {
             return Err(TokenError::InvalidNegative(format!(
                 "mint amount {} cannot be negative",
                 value
             )));
         }
 
+        let old_state = self.state.clone();
```
anorth, line 114: Note: It's slightly sad that this clone followed by the transaction results in a redundant clone, and that we do an actual write to rollback when the old state is often already persisted. The state is small so it doesn't matter _too_ much. 

An alternative pattern would involve keeping track of the CID at load, and reverting by re-loading it. It this presents a snapshot (could avoid all the clones this way, but I'm not sure which is cheaper). It doesn't immediately apply here tho since we don't know whether the prior state has been modified since load.

fil_fungible_token/src/token/mod.rs
```diff
@@ -80,28 +89,64 @@ where
...
-        self.state.change_balance_by(&self.bs, initial_holder, &value)?;
+        self.transaction(|state, bs| {
+            state.change_balance_by(&bs, initial_holder, value)?;
+            state.change_supply_by(value)?;
+            Ok(())
+        })?;
 
-        // TODO: invoke the receiver hook on the initial_holder
+        // Update state so re-entrant calls see the changes
+        self.flush()?;
```
anorth, line 124: This might flush previously un-flushed state changes. We do need to flush here, but this could surprise someone so it's pointing to something just slightly off in APIs somewhere.

Basically there's a conflict between the need to flush here, and objectives about holding up state changes for batched flush. Maybe something like this is the best compromise available. I can't see how to sensibly avoid a flush between every transfer or mint, even if doing a large batch of them. But then maybe for consistency we give up on the idea of non-persisted changes at the level of Token API methods.

fil_fungible_token/src/token/mod.rs
anorth: Yes, see above

fil_fungible_token/src/token/mod.rs
```diff
@@ -272,71 +318,92 @@ where
...
+                    by: spender,
+                    value: value.clone(),
+                })
+            }
+            Err(e) => {
+                // error calling receiver hook, revert state
+                self.state = old_state;
+                self.flush()?;
+                Err(e.into())
+            }
```
anorth, line 358: Is it possible to factor out this repeated code?

fil_fungible_token/src/token/mod.rs
```diff
@@ -376,15 +443,32 @@ mod test {
...
 
+    #[test]
+    fn it_fails_to_mint_if_receiver_hook_aborts() {
+        let mut token = new_token();
+
+        let treasury = 1;
+        // force hook to abort
+        token
+            .mint(TOKEN_ACTOR_ADDRESS, treasury, &TokenAmount::from(1_000_000), "abort".as_bytes())
+            .unwrap_err();
```
anorth, line 454: As noted above, I'd like to see tests that the hook was invoked with expected parameters, i.e. with a mock.

fil_fungible_token/src/method.rs
alexytsu: https://github.com/helix-onchain/filecoin/issues/31

fil_fungible_token/src/method.rs
alexytsu: https://github.com/helix-onchain/filecoin/issues/33

fil_fungible_token/src/token/mod.rs
alexytsu: image

A quick test using the fvm_integration_tests package seems to show that the clone and revert strategy is cheaper than reloading the old state from the blockstore

Since the root level of state is just a BigInt and 2 CIDs, the only real difference is whether we revert those three values from an in-memory copy (should be cheap) or a copy in the blockstore. In fact, the testing I did seems to suggest that there's no gas difference in making that clone (but not sure if there's something wrong with the testing methodology here).

fil_fungible_token/src/token/mod.rs
alexytsu: I'll defer this refactor after some more benchmarking to settle on the preferred rollback strategy

fil_fungible_token/src/token/mod.rs
anorth: Thanks, I'm happy for us to roll with clones then.

anorth (approved):

PR #29, opened 2022-07-22
run PR checks with code coverage

alexytsu:

PR #34, opened 2022-07-25
Method hash macro

abright:
Adds a procedural macro to do hashing of method names at compile time. For convenience it can be fed a string or a function name, with function names being converted from snake_case to PascalCase. Hashing is done using the `MethodResolver` borrowed from `fvm_dispatch` so naming rules are also enforced during compilation.

It currently has its own copy of the `MethodResolver` code because I can't pull in the entire `fvm_dispatch` crate. There's a link error when I try doing that but I believe it's off in another dependency somewhere, so avoiding all the extra imports is the safest way to go (also, helps reduce the impact on build times). I want everything to use the same code for validating names and generating the hashes and there should be a couple ways to go about it:
- Extract the `MethodResolver` / hashing code into its own package
- Add feature flags to its components so it only builds/imports what it needs (might be a better solution, but might also lead down a rabbit hole of ever more flags and fine-grained dependency management)

TODO:
- [x] Use the same `MethodResolver` / hashing code as `fvm_dispatch`, to avoid maintaining a local copy
- [x] Proper tests beyond the barebones example, especially around the naming rules. Will put something together using `trybuild` for this
- [x] Update package name/location. ~It's somewhat separate to `fvm_dispatch` and can't use that entire crate (due to linking issues mentioned above) so should probably stand alone.~

Update: Might still want to rename this from "helper" to something else? But it's now using `MethodResolver` from `fvm_dispatch` instead of maintaining its own copy


abright:
Looks like https://github.com/helix-onchain/filecoin/pull/29 hits the same (or similar) issue that I had with depending on `fvm_sdk`

fvm_dispatch/helper/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            Err(lookahead.error())
+        }
+    }
+}
+
+#[proc_macro]
+pub fn method_hash(input: TokenStream) -> TokenStream {
+    let name: MethodName = parse_macro_input!(input);
+    let hash = name.hash() as u32;
```
abright, line 49: Should this output a `u32` or a `u64`?

anorth:
Cool, this is a good start.

I would like to simplify as much as possible, and to that end I think we should drop the handling of tokens as input (vs strings). I think it will be more robust to require devs to type their exported method name string, and have the method it's dispatched to be able to change independently. I know they _can_ do that, but for simplicity now I think we should require it. My view could change upon seeing how this integrates into a higher level dispatch table macro, though.

I'm also skeptical that the case library will do a good job. E.g. acronyms should be capitalised. I think forcing human intervention here will lead to better programs.

anorth:
Re deps, I would have suggested we just put it in the existing dispatch package. Too many crates will add dependency pain over time. The dispatch crate is already pretty small.

Second to that I would have said just depend on the existing dispatch package.

I'm very hesitant to go down the road of build-time feature flags, for the reasons you suggest.

abright:
> Re deps, I would have suggested we just put it in the existing dispatch package. Too many crates will add dependency pain over time. The dispatch crate is already pretty small.
> 
> Second to that I would have said just depend on the existing dispatch package.
> 
> I'm very hesitant to go down the road of build-time feature flags, for the reasons you suggest.

Proc macros do need to live in a separate package, because they're built and run as part of the build process. The linking issue I had when depending on `fvm_dispatch` seems similar to what I saw in https://github.com/helix-onchain/filecoin/pull/29 but there might be an easy way to deal with it here. There's a config flag set when building a proc macro package which I should be able to use to block the `fvm_sdk` related parts of `fvm_dispatch` as the proc macro build doesn't need them. There's also [`target_family = "wasm"`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_family) which might work, but I'll try to keep everything specific to my proc macro build for now.

abright:
Seems the `#[cfg(proc_macro)]` flag doesn't do much(anything?) when it comes to dependencies, so I've used `#[cfg(target_family = "wasm")]` instead. When building for WASM, the `fvm_sdk` syscalls will work as expected, but they'll be replaced with stubs that return an error (call_method) or empty (hash) result to allow non-WASM builds to use the package.

Ident hashing is removed also, so it's purely string input now.

anorth:
> Proc macros do need to live in a separate package, because they're built and run as part of the build process.

Ok thanks for explaining.

Re linking, it would be great obviously to fix the underlying issue. `[cfg(target_family = "wasm")]` looks ok to get you unblocked now, but I worry it'll cause problems later if it's kinda outside the builders control. E.g. how about when building integration tests against a framework that does use the full FVM SDK (because it patches lower level at the syscalls). Would an explicit feature flag work here, that the builder can toggle (or by default follows the target family)?

@alexytsu this is another data point around the general problem of testability of things that invoke syscalls. I _think_that we might want a shim at the very lowest syscall layer, but the SDK isn't set up for that right now. So we have to shim the SDK itself instead. I guess that's ok and is more or less what's happening here. Maybe nothing for us to improve at the moment here.

abright:
> I _think_that we might want a shim at the very lowest syscall layer, but the SDK isn't set up for that right now.

A couple ideas I've had along those lines:

-  Build a standalone package by copying the `sys` code out of the SDK and modify the `fvm_syscalls` macro to generate stubs for everything. Adds some maintenance burden (to keep in sync with the SDK whenever the API changes) but it's a drop-in fix to resolve the `extern` functions at link time so users could just tell cargo to use it for non-wasm targets, or add it to dev-dependencies or whatever. Doesn't require changing SDK or user code (and would allow me to remove those shims in here) so it might be a good short- to medium-term solution?
-  Doing a similar thing inside the SDK by using feature or target flags to choose between modules containing the real or stub versions of the macro could be a more complete solution. Some crates do things like this for platform-specific implementations of a particular feature. Should be almost maintenance-free once completed but the selection mechanism could get a bit too magical.

I guess that's also heading off-topic for this PR though. I think this is good to go unless you want me to change the package name or anything like that?

abright:
Looks like the build flags added for CI in PR #29 aren't playing nice with something here due to different panic handlers. Will look into it.

fvm_dispatch/helper/example/src/main.rs
```diff
@@ -0,0 +1,11 @@
+use helper::method_hash;
+
+fn main() {
+    // these should produce an identical result
+    // because identifiers are converted to PascalCase for hashing
```
anorth, line 5: Remove stale comment.

fvm_dispatch/helper/src/lib.rs
anorth: The `MethodNum` type in `fvm_shared` is `u64`.

fvm_dispatch/helper/tests/build-success.rs
```diff
@@ -0,0 +1,8 @@
+use helper::method_hash;
+
+fn main() {
+    // these should produce identical output
```
anorth, line 4: Stale comment

fvm_dispatch/src/hash.rs
```diff
@@ -16,9 +17,18 @@ pub trait Hasher {
 pub struct Blake2bSyscall {}
 
 impl Hasher for Blake2bSyscall {
+    // fvm_sdk syscalls only work for WASM targets
+    #[cfg(target_family = "wasm")]
```
anorth, line 21: Do we still need this? I confess I don't fully understand how it will work. What will happen when someone wants to use some new FVM integration test thing that runs purely native (no WASM)?

anorth (approved):
Let's change the package name, "helper" is too generic. We could use "macros", especially if we'll add other supporting macros here. "Support" could work?

fvm_dispatch/src/hash.rs
abright: For the moment we need it, to avoid unresolved externs (part of the syscalls within the SDK). We might be able to get rid of it if using the FVM integration test library, I haven't looked closely at that code yet. I've changed this (and the method call one in `message.rs`) to use `unimplemented!()` which will panic if they're ever called instead of returning empty value or error results.

Added an issue to address it later: https://github.com/helix-onchain/filecoin/issues/43

PR #37, opened 2022-07-27
Token transaction builder API

alexytsu:
A WIP transaction-based API. This uses a "builder" style pattern of method chaining to perform many fallible operations in sequence. If at any point in the chain a method fails, subsequent operations are skipped and the final `flush` reverts the Token state.

There's quite a bit of additional complexity from attempting to have `DirtyTransaction` and `CleanTransaction` be interoperable. It would be nice to have `mint` and `transfer` preconditions enforced at compile time but I haven't quite gotten it working and it results in extra clones. @abright would be nice to discuss with you if there's a way to restructure the traits and structs to achieve this efficiently. 

I think an alternate API would be better where we have a single Transaction interface that includes `mint` and `transfer` functionality. The checking would be done at runtime (with a "dirty-state" flag). This would allow our mutation methods to return `&mut Transaction` rather than the current API which requires returning a new instance of a `DirtyTransaction`. The current limitation is required as `CleanTransaction` also needs to be able return a `DirtyTransaction`.

### DIFF
--- a/fil_fungible_token/src/blockstore.rs
+++ b/fil_fungible_token/src/blockstore.rs
@@ -17,9 +17,7 @@ pub struct Blockstore;
 impl fvm_ipld_blockstore::Blockstore for Blockstore {
     fn get(&self, cid: &Cid) -> Result<Option<Vec<u8>>> {
         // If this fails, the _CID_ is invalid. I.e., we have a bug.
-        ipld::get(cid)
-            .map(Some)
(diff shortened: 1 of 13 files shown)


anorth:
Thanks. I agree this is probably more fiddly than it's worth

alexytsu:
Simplified spike here
https://github.com/helix-onchain/filecoin/pull/40

PR #38, opened 2022-07-28
run coverage reports on main

alexytsu:


abright (approved):
Is this just an informative thing or will we use it to enforce a minimum coverage level?

alexytsu:
Informative for now, might be worth enforcing some rules on PRs once the development has stabilised a bit.

Most likely to prevent coverage from regressing, wdyt?

abright:
> 
> Most likely to prevent coverage from regressing, wdyt?

Sounds good to me

PR #44, opened 2022-08-01
Token handle refactor

alexytsu:
Having run into a few issues with usability of the library the following restructuring has been made

- add resolve_to_id_addr utility (moved to https://github.com/helix-onchain/filecoin/pull/46)
- allow Token to wrap an already loaded state
- invert hierarchy of handle and state object
- do not allow mutation via a raw TokenState

Also have managed to avoid cloning the blockstore after studying the builtin-actor's runtime usage


fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -32,12 +35,18 @@ pub trait MethodCaller {
         token_value: &TokenAmount,
         data: &[u8],
     ) -> Result<Receipt>;
+
+    /// Resolves the given address to it's ID address form
+    ///
+    /// If an ID address for the given address doesn't exist yet, it tries to create one by sending a
+    /// zero balance to the given address.
+    fn resolve_to_id_addr(&self, address: &Address) -> Result<ActorID>;
```
anorth, line 43: On balance I think it would be better if this only did the resolution, and the send-to-initialise was encapsulated in the token library (and testable) rather than injected by the user.

anorth (commented):
Thanks. Can you break this address resolution out to its own PR? It'll be good to land quickly.

Please add tests for the token library handling un-resolvable addresses.

fil_fungible_token/src/token/mod.rs
```diff
@@ -45,40 +45,36 @@ type Result<T> = std::result::Result<T, TokenError>;
...
+        Ok(token)
     }
 
-    /// For an already initialised state tree, loads the state tree from the blockstore and returns
-    /// a Token handle to interact with it
-    pub fn load(bs: BS, msg: MSG, state_cid: Cid) -> Result<Self> {
-        let state = TokenState::load(&bs, &state_cid)?;
-        Ok(Self { bs, msg, state })
+    /// Returns a mutable reference to underlying token-state
+    pub fn state(&mut self) -> &mut State {
```
anorth: At a first guess, I would think we don't want this to be mutable.

anorth: Thanks for following my quick hacks here. I think we can find better/safer flows. There are two main use cases I'm aware of right now, and they might each benefit from a helper.
1. Token state is the only state for an actor, or linked by CID. In this case we want to retain something like the old init_state path that creates state, and there's no need to expose it.
2. Token state is embedded in actor state. In this case we need to wrap, and this `state` method I added was to let the caller copy the state back out. But that's an avenue for error (forgetting to do that). If it's a mutable reference, maybe we don't need this at all? Another alternative is a `with(state, callback(token))` transaction-style method that exposes the Token API only within a closure, and ensures writeback.

Something like the callback method could work for (1) too.

Another alternative for (1) would be a `Rc` over the state, but we can probably avoid that.

I think it'll be very hard to get these APIs right without working an example actor of each type.

fil_fungible_token/src/token/mod.rs
```diff
@@ -45,40 +45,36 @@ type Result<T> = std::result::Result<T, TokenError>;
...
 }
 
-impl<BS, MSG> Token<BS, MSG>
+impl<'st, BS, MSG> Token<'st, BS, MSG>
 where
-    BS: IpldStore + Clone,
+    BS: IpldStore,
     MSG: Messaging,
 {
     /// Creates a new token instance using the given blockstore and creates a new empty state tree
```
anorth: Out of date

anorth:
I do not understand what's going on with the invert hierarchy change, sorry. Could you perhaps undo moving the files so that a diff is more approachable?

alexytsu:
> I do not understand what's going on with the invert hierarchy change, sorry. Could you perhaps undo moving the files so that a diff is more approachable?

The idea here is to enforcing mutating state only via the `Token` handle and prevent direct mutations to state via the `TokenState` struct. So I've made those methods private on `TokenState` but for according to Rust visibility rules, for `Token` to be able to call them, it must be a submodule of `TokenState`

abright:
You might be able to avoid the path changes if pub visibility was the only reason to move them: https://doc.rust-lang.org/reference/visibility-and-privacy.html - maybe `pub(crate)` or `pub(super)` could do the job

fil_fungible_token/src/runtime/messaging.rs
alexytsu: https://github.com/helix-onchain/filecoin/pull/46/files

New structure in here allows more testing of the behaviour on uninitialized accounts

alexytsu:
this has been broken down into smaller PRs

Issue #33, opened 2022-07-25
Mock a dynamic MethodCaller which performs verification for testing


alexytsu:

alexytsu:
https://github.com/helix-onchain/filecoin/pull/47 improves the utility of `FakeMessenger` to mock more behaviours and allows inspection of sent messages to verify their shape

PR #46, opened 2022-08-02
Replaces ActorIDs with Addresses where appropriate

alexytsu:


alexytsu:
I'm unsure if `readonly` methods such as `balance_of` `allowance` should cause uninitialized accounts to be created

anorth:
> I'm unsure if readonly methods such as balance_of allowance should cause uninitialized accounts to be created

No I think not.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// Returns true if the receiver hook is called and exits without error, else returns false
+    fn call_receiver_hook(
+        &self,
+        from: ActorID,
+        to: ActorID,
+        token_value: &TokenAmount,
+        data: &[u8],
+    ) -> Result<Receipt>;
+
+    /// Resolves the given address to it's ID address form
```
anorth, line 39:
```suggestion
    /// Resolves the given address to its ID address form
```

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+        // TODO: ^^ necessitates determining conventional method names for receiver hooks
+
+        // currently, the method number comes from taking the name as "TokensReceived" and applying
+        // the transformation described in https://github.com/filecoin-project/FIPs/pull/399
+        const METHOD_NUM: u64 = 1361519036;
+        let to = Address::new_id(to);
+
+        let params = TokenReceivedParams {
+            sender: Address::new_id(from),
+            value: token_value.clone(),
```
anorth, line 67: Nit: I think I'd like to rename `value` to `amount` everywhere, as in "amount of tokens". The VM will continue to use `value` for native tokens only.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+            data: RawBytes::from(data.to_vec()),
```
anorth, line 68: For a follow-up PR, replace `&[u8]` with RawBytes everywhere it's used as receiver data, which will remove some noise here and elsewhere.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn call_receiver_hook(
+        &self,
+        from: ActorID,
+        to: ActorID,
+        token_value: &TokenAmount,
+        data: &[u8],
+    ) -> Result<Receipt>;
+
+    /// Resolves the given address to it's ID address form
+    fn resolve_id(&self, address: &Address) -> Result<ActorID>;
```
anorth, line 40: I think it's worth documenting the concrete "error" returned when the address is not resolvable, so that caller can catch that one explicitly while passing on others. Ditto below.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+/// If call_receiver_hook is called with an empty data array, it will return true.
+/// If call_receiver_hook is called with a non-empty data array, it will return false.
+///
+/// # resolve_id
+/// If Address is id-address or secp-address, it resolves by returing FAKE_RESOLVED_ID
+/// If Address is actor-address or bls-address, it returns an error simulating an uninitialised address
+///
+/// # initialise_account
+/// Simulates the initialisation of an account
+/// - ID, SECP addresses panic as they should not be called given that resolve_id gave back an ActorID
```
anorth, line 100: In the real implementation, ID, SECP, and BLS should all succeed always, and Actor always fail. I think we should replicate that here because callers may depend on that behaviour.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+/// - BLS addresses give back FAKE_INITIALIZED_ID
+/// - Actor addresses are uninitialised and give back an error
+#[derive(Debug, Default, Clone, Copy)]
+pub struct FakeMessenger {}
+
+pub const FAKE_RESOLVED_ID: ActorID = 100;
+pub const FAKE_INITIALIZED_ID: ActorID = 101;
```
anorth, line 107: Probably not an issue right now, but really these should be variables so we can return distinct ids for different addresses. In fact they should be a map of already resolved ones so they are stable. This will probably bite someone at some point.  You can defer to #33 if you like.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// A fake method caller
+///
+/// # call_receiver_hook
+/// If call_receiver_hook is called with an empty data array, it will return true.
+/// If call_receiver_hook is called with a non-empty data array, it will return false.
+///
+/// # resolve_id
+/// If Address is id-address or secp-address, it resolves by returing FAKE_RESOLVED_ID
+/// If Address is actor-address or bls-address, it returns an error simulating an uninitialised address
```
anorth, line 96: I feel like there will make test-writing a little fiddly, and cause wide churn if we ever want to change these behaviours. All pushing towards either a much more robust fake, that can be configured, or a mock (#33).

fil_fungible_token/src/token/mod.rs
```diff
@@ -101,20 +102,38 @@ where
...
+    MSG: Messaging,
 {
     /// Mints the specified value of tokens into an account
     ///
-    /// The mint amount must be non-negative or the method returns an error
+    /// The minter is implicitly defined as the caller of the actor.
+    /// The mint amount must be non-negative or the method returns an error.
     pub fn mint(
         &mut self,
         minter: ActorID,
```
anorth, line 135: Even though we know it will be an ID address, take `Address` here because that's what the runtime gives us. Put the `.id().map_err()` noise inside the token library rather than require it at call sites.

fil_fungible_token/src/token/mod.rs
```diff
@@ -176,28 +197,32 @@ where
     /// Returns the balance associated with a particular address
     ///
     /// Accounts that have never received transfers implicitly have a zero-balance
-    pub fn balance_of(&self, holder: ActorID) -> Result<TokenAmount> {
+    pub fn balance_of(&self, holder: &Address) -> Result<TokenAmount> {
         // Load the HAMT holding balances
+        let holder = self.resolve_to_id_addr(holder)?;
```
anorth, line 202: Use get_id_addr here – we don't want to create the actor. If it doesn't have an ID address it can't have a balance.

fil_fungible_token/src/token/mod.rs
```diff
@@ -176,28 +197,32 @@ where
...
     }
 
     /// Gets the allowance between owner and spender
     ///
     /// The allowance is the amount that the spender can transfer or burn out of the owner's account
     /// via the `transfer_from` and `burn_from` methods.
-    pub fn allowance(&self, owner: ActorID, spender: ActorID) -> Result<TokenAmount> {
+    pub fn allowance(&self, owner: &Address, spender: &Address) -> Result<TokenAmount> {
+        let owner = self.resolve_to_id_addr(owner)?;
+        let spender = self.resolve_to_id_addr(spender)?;
```
anorth, line 212: `get_id_addr` both of these for similar reason. increase_allowance will have initialised if it was called, otherwise allowance must be zero.

fil_fungible_token/src/token/mod.rs
```diff
@@ -176,28 +197,32 @@ where
...
     }
 
     /// Increase the allowance that a spender controls of the owner's balance by the requested delta
     ///
+    /// The caller of this method is implicitly defined as the owner.
     /// Returns an error if requested delta is negative or there are errors in (de)serialization of
     /// state. Else returns the new allowance.
     pub fn increase_allowance(
         &mut self,
         owner: ActorID,
```
anorth, line 224: Address here and decrease_allowance too

fil_fungible_token/src/token/mod.rs
```diff
@@ -230,14 +257,16 @@ where
...
         let new_allowance =
             self.state.change_allowance_by(&self.bs, owner, spender, &delta.neg())?;
 
         Ok(new_allowance)
     }
 
     /// Sets the allowance between owner and spender to 0
-    pub fn revoke_allowance(&mut self, owner: ActorID, spender: ActorID) -> Result<()> {
+    pub fn revoke_allowance(&mut self, owner: ActorID, spender: &Address) -> Result<()> {
+        let spender = self.resolve_to_id_addr(spender)?;
```
anorth, line 269: `get_id_addr`

fil_fungible_token/src/token/mod.rs
```diff
@@ -101,20 +102,38 @@ where
...
+    fn resolve_to_id_addr(&self, initial_holder: &Address) -> Result<ActorID> {
+        let holder_id = match self.msg.resolve_id(initial_holder) {
+            Ok(addr) => addr,
+            Err(_) => self.msg.initialise_account(initial_holder)?,
+        };
+        Ok(holder_id)
+    }
+
+    /// Attempts to resolve an address to an ID address, returning an error if it did not exist
+    fn get_id_addr(&self, initial_holder: &Address) -> Result<ActorID> {
```
anorth, line 119: Nit: perhaps drop `addr` from the end of these names since they return IDs.
- resolve_addr_id
- get_addr_id
?

fil_fungible_token/src/token/mod.rs
```diff
@@ -269,7 +298,7 @@ where
     pub fn burn(
         &mut self,
         spender: ActorID,
```
anorth, line 300: Address

fil_fungible_token/src/token/mod.rs
```diff
@@ -322,8 +354,8 @@ where
     pub fn transfer(
         &mut self,
         spender: ActorID,
```
anorth, line 356: Address here too

anorth (commented):
Thanks this is great, just needs a bit more!

alexytsu:
@anorth this is ready for re-review
I'm deferring https://github.com/helix-onchain/filecoin/pull/46#discussion_r936132262
and https://github.com/helix-onchain/filecoin/pull/46#discussion_r936131896
to followups

anorth (approved):
Approving so we can roll forward, this is basically right.

There's still something I think we need to figure out about the intersections of uninitialized accounts and zero transfers. Let's discuss it in Slack.

PR #47, opened 2022-08-02
fix mint receiver hook

alexytsu:
Construction of the TokenReceiverHook params is now performed in the Token library rather than in the injected layer and tested

### DIFF
--- a/fil_fungible_token/src/receiver/types.rs
+++ b/fil_fungible_token/src/receiver/types.rs
@@ -1,7 +1,8 @@
 use fvm_ipld_encoding::tuple::{Deserialize_tuple, Serialize_tuple};
 use fvm_ipld_encoding::{Cbor, RawBytes};
 use fvm_shared::bigint::bigint_ser;
-use fvm_shared::{address::Address, econ::TokenAmount};
+use fvm_shared::econ::TokenAmount;
+use fvm_shared::ActorID;
 
 /// Standard interface for an actor that wishes to receive FRC-XXX tokens
 pub trait FrcXXXTokenReceiver {
@@ -15,9 +16,14 @@ pub trait FrcXXXTokenReceiver {
     fn token_received(params: TokenReceivedParams);
 }
 
-#[derive(Serialize_tuple, Deserialize_tuple)]
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
 pub struct TokenReceivedParams {
-    pub sender: Address,
+    /// Address of the sender or operator that initiated the transfer
+    pub sender: ActorID,
+    /// The account that the tokens are being pulled from (the token actor address itself for mint)
+    pub from: ActorID,
+    /// The account that the tokens are being sent to (the receiver address)
+    pub to: ActorID,
     #[serde(with = "bigint_ser")]
     pub value: TokenAmount,
     pub data: RawBytes,
--- a/fil_fungible_token/src/runtime/messaging.rs
+++ b/fil_fungible_token/src/runtime/messaging.rs
@@ -3,9 +3,10 @@ use std::collections::HashMap;
 
 use fvm_ipld_encoding::Error as IpldError;
 use fvm_ipld_encoding::RawBytes;
-use fvm_sdk::{actor, send, sys::ErrorNumber};
+use fvm_sdk::{actor, message, send, sys::ErrorNumber};
 use fvm_shared::error::ExitCode;
 use fvm_shared::receipt::Receipt;
+use fvm_shared::MethodNum;
 use fvm_shared::METHOD_SEND;
 use fvm_shared::{address::Address, econ::TokenAmount, ActorID};
 use num_traits::Zero;
@@ -27,11 +28,17 @@ pub enum MessagingError {
 
 /// An abstraction used to send messages to other actors
 pub trait Messaging {
-    /// Call the receiver hook on a given actor, specifying the amount of tokens pending to be sent
-    /// and the sender and receiver
-    ///
-    /// Returns true if the receiver hook is called and exits without error, else returns false
-    fn call_receiver_hook(&self, to: &Address, params: &TokenReceivedParams) -> Result<Receipt>;
+    /// Returns the address of the current actor as an ActorID
+    fn actor_id(&self) -> ActorID;
+
+    /// Sends a message to an actor
+    fn send(
+        &self,
+        to: &Address,
+        method: MethodNum,
+        params: &RawBytes,
+        value: &TokenAmount,
+    ) -> Result<Receipt>;
 
     /// Resolves the given address to its ID address form
     ///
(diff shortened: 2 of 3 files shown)


fil_fungible_token/src/receiver/types.rs
```diff
@@ -15,9 +16,12 @@ pub trait FrcXXXTokenReceiver {
...
 }
 
-#[derive(Serialize_tuple, Deserialize_tuple)]
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
 pub struct TokenReceivedParams {
-    pub sender: Address,
+    /// Address of the sender or operator that initiated the transfer
+    pub sender: ActorID,
+    /// The account that the tokens are being pulled from (the token actor address itself for mint)
+    pub from: ActorID,
```
anorth, line 24: It's an interesting question whether these values should be ActorId or Address. Certainly, if Address, it should be an ID type address. And so passing ID here makes that very clear. But everywhere else (e.g. syscalls, built-in actors) always pass addresses. Perhaps for FIP discussion.

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -47,29 +43,17 @@ pub trait Messaging {
...
 
-        let params = TokenReceivedParams {
-            sender: Address::new_id(from),
-            value: token_value.clone(),
-            data: RawBytes::from(data.to_vec()),
-        };
         let params = RawBytes::new(fvm_ipld_encoding::to_vec(&params)?);
 
-        Ok(send::send(&to, METHOD_NUM, params, TokenAmount::zero())?)
+        Ok(send::send(to, METHOD_NUM, params, TokenAmount::zero())?)
```
anorth, line 56: Now that we're here, maybe even METHOD_NUM should be pushed out so that this _only_ wraps the SDK method. This could then form a generic interface that would be useful for other libraries too.

fil_fungible_token/src/token/mod.rs
```diff
@@ -57,6 +59,8 @@ where
     msg: MSG,
     /// In-memory cache of the state tree
     state: TokenState,
+    /// ActorID
```
anorth, line 62: Receiving actor ID, i.e. `message().receiver()`

But also why not make this a function of MSG? It's also an abstraction from the syscall interface

fil_fungible_token/src/receiver/types.rs
anorth: A new thought: I think we should add the `to` here as well. Yes the receiver will know their own address, but including it explicitly will allow the message to be transparently forwarded, used with proxy patterns etc. ERC-777 includes it (where delegate calls etc make the need more obvious). FVM in the future might add delegation capabilites that make this relevant.

anorth (approved):
Approving so you can either merge and follow-up, or take suggestions in this PR.

alexytsu:
I rebased this off of the updated https://github.com/helix-onchain/filecoin/pull/46 and made some of the suggested changes

For now, the hook still takes `ActorIDs` rather than `Addresses` but I am keeping an eye on https://github.com/filecoin-project/FIPs/discussions/407#discussioncomment-3312126 or other discussions around this

fil_fungible_token/src/receiver/types.rs
```diff
@@ -15,9 +16,14 @@ pub trait FrcXXXTokenReceiver {
     fn token_received(params: TokenReceivedParams);
 }
 
-#[derive(Serialize_tuple, Deserialize_tuple)]
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
 pub struct TokenReceivedParams {
-    pub sender: Address,
+    /// Address of the sender or operator that initiated the transfer
+    pub sender: ActorID,
```
anorth, line 22: Discussion: in this context the term "sender" is easily confusable with "from". This makes me think we should adopt the term "operator" throughout (with a slightly different meaning that ERC-777).

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -94,56 +91,64 @@ impl Messaging for FvmMessenger {
...
-                gas_used: 0,
-            })
+                return_data: Default::default(),
+            });
+        }
+
+        if method == RECEIVER_HOOK_METHOD_NUM {
+            let hook_params = params.deserialize().unwrap();
+            self.last_hook.borrow_mut().replace(hook_params);
         }
```
anorth, line 149: else fail? whatever method num was called probably doesn't exist (until we add more mocking ability)

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -94,56 +91,64 @@ impl Messaging for FvmMessenger {
...
-                return_data: RawBytes::new(data.to_vec()),
-                gas_used: 0,
-            })
+                return_data: Default::default(),
+            });
+        }
+
+        if method == RECEIVER_HOOK_METHOD_NUM {
+            let hook_params = params.deserialize().unwrap();
+            self.last_hook.borrow_mut().replace(hook_params);
```
anorth, line 148: Can we do this before abort_next_send so we can still inspect the params?

fil_fungible_token/src/token/mod.rs
```diff
@@ -174,7 +170,19 @@ where
         self.flush()?;
 
         // Call receiver hook
-        match self.msg.call_receiver_hook(minter, holder_id, value, data) {
+        let hook_params = TokenReceivedParams {
+            data: RawBytes::from(data.to_vec()),
```
anorth, line 174: Propagate the `RawBytes` change up to the `data` param too

fil_fungible_token/src/token/mod.rs
```diff
@@ -415,24 +423,36 @@ where
         // spender must be an id address
         let spender = expect_id(spender)?;
         // resolve owner and receiver
-        let owner = self.resolve_to_id(owner)?;
-        let receiver = self.resolve_to_id(receiver)?;
+        let owner_id = self.resolve_to_id(owner)?;
+        let receiver_id = self.resolve_to_id(receiver)?;
```
anorth, line 427: Nit: just because "receiver" is overloaded by the syscalls to refer to the message receiver (in the OO sense, i.e. currently executing actor), let's use a different word here. Either `from`/`to` consistently is probably best, even if awkward English. Can defer to follow-up

fil_fungible_token/src/token/mod.rs
```diff
@@ -415,24 +423,36 @@ where
...
         self.flush()?;
 
         // call receiver hook
-        match self.msg.call_receiver_hook(owner, receiver, value, data) {
+        let params = TokenReceivedParams {
+            sender: spender,
+            from: owner_id,
+            to: receiver_id,
+            value: value.clone(),
+            data: RawBytes::new(data.to_vec()),
```
anorth, line 448: Ditto `data: RawBytes`

anorth (approved):
Great. Mior suggestions, and some terminology to clean up (maybe later). I refrained from commenting on any address resolution stuff we've been discussing in Slack.

PR #49, opened 2022-08-03
Don't require a clonable blockstore

alexytsu:
do not require a clonable blockstore


anorth (approved):
Awesome.

anorth:
@zenground0 @stebalien might this be possible in the built-in actors too?

PR #51, opened 2022-08-04
Re-organise dispatch and macro crates

abright:
Noisy diffs in this one (but the commits cover one change each and will be easier to review) because it moves and renames a bunch of things.

- hashing moved out into its own (sub-)crate
- proc macro crate also renamed to become a (sub-)crate
- macro and hashing stuff transparently re-exported through `fvm_dispatch`

This paves the way for declarative (`macro_rules!`) and proc macros to (appear to) live in the same crate. Also improves build times somewhat as the macro crate depends on the hashing crate directly instead of the full `fvm_dispatch`, and avoids pulling in `fvm_sdk` entirely with a `no_sdk` feature flag.

Anything using the hashing functions from `fvm_dispatch` will continue to work as before, while the method hashing macro should now be available as `fvm_dispatch::method_hash`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 
 members = [
     "fvm_dispatch",
+    "fvm_dispatch/hasher",
     "fvm_dispatch/macros",
     "fvm_dispatch/macros/example",
     "fvm_dispatch_tools",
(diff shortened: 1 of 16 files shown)


anorth (commented):
Cool, look forward to seeing it in action in example actors

anorth (approved):

abright:
PR for the (WIP) dispatch macro: https://github.com/helix-onchain/filecoin/pull/52

PR #52, opened 2022-08-04
Method match macro and example actor

abright:
Adds a macro to generate a `match` for method dispatching in an actor's `invoke` function along with a simple (but complete) example that can be run using `fvm_integration_tests`

```Rust
fn invoke(input: u32) -> u32 {
    let method_num = sdk::message::method_number();
    match_method!(method_num, {
        "Constructor" => {
            // this is a stateless actor so constructor does nothing
            NO_DATA_BLOCK_ID
        },
        "Greet" => {
            // TODO: load params, call greet(), save result and return block id
        },
        _ => {
            // abort due to unknown method
        }
    })
}
```

currently expands to this at build time (see the greeter example for the full body code)
```Rust
fn invoke(input: u32) -> u32 {
    let method_num = sdk::message::method_number();
    match method_num {
        1u64 => NO_DATA_BLOCK_ID,
        225812285u64 => {
            // TODO: load params, call greet(), save result and return block id
        }
        _ => {
            // abort due to unknown method
        }
    }
}
```

### DIFF
--- a/.github/workflows/check-main.yml
+++ b/.github/workflows/check-main.yml
@@ -15,8 +15,11 @@ jobs:
       - name: Build
         run: make check-build
 
-      - name: Test
-        run: make test
+      - name: Test Actors
+        run: make test-actors
+
+      - name: Test Libraries
+        run: make test-coverage
         env:
           CARGO_INCREMENTAL: '0'
           RUSTFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Coverflow-checks=off'
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -20,8 +20,11 @@ jobs:
       - name: Build
         run: make check-build
 
-      - name: Test
-        run: make test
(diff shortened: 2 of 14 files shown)


abright:
A recent change to Rust's standard library broke builds of this when using the latest nightly toolchain: https://github.com/rust-lang/rust/commit/b2bbca393384c6d7e343decef5a927e16f003038 due to `fvm` using the `backtrace` function that was removed. 

As a workaround, I've changed our toolchain override to use the same version that the `ref-fvm` repo does. This would also fix the CI failure in https://github.com/helix-onchain/filecoin/pull/55

abright:
Current CI failure seems to be caused by profiling support (used by test coverage, I think) not being available when building for WASM targets. Will need to find a workaround for this to keep CI happy but the code will build and run locally.

anorth (approved):
I feel underqualified to review the macro code itself, but the result looks like a good step.

alexytsu (approved):
Reviewed mostly the CI and general structure which LGTM

PR #53, opened 2022-08-04
Upgrade to nightly toolchain

alexytsu:
This is needed in order to compile example actors

It also happens to fix the build issues in CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,19 @@
-build:
+install-toolchain:
+	rustup component add rustfmt
+	rustup component add clippy
+	rustup target add wasm32-unknown-unknown
+
+build: install-toolchain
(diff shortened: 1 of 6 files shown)


abright (approved):
LGTM

Was there some specific feature/issue that required nightly? If so, we can keep an eye on when that stuff makes it into stable

alexytsu:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
   -->.cargo/registry/src/github.com-1ecc6299db9ec823/bellperson-0.21.0/src/lib.rs:139:46
    |
139 | #![cfg_attr(target_arch = "aarch64", feature(stdsimd))]
```

This was the immediate blocker but there may be others

PR #54, opened 2022-08-04
Parameter renaming

alexytsu:
- rename value -> amount when referring to tokens https://github.com/helix-onchain/filecoin/pull/46#discussion_r936131896
- replace &[u8] with RawBytes https://github.com/helix-onchain/filecoin/pull/46#discussion_r936132262

Third commit (maybe too drastic) unifies terminology for sender, spender, minter -> operator. The term operator captures the actor or entity that initiates a state change (or has the power to in the case of allowances)

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -28,11 +28,14 @@ pub enum TokenError {
     InvalidNegative(String),
     #[error("error calling receiver hook: {0}")]
     Messaging(#[from] MessagingError),
-    #[error("receiver hook aborted when {from:?} sent {amount:?} to {to:?} by {sender:?} with exit code {exit_code:?}")]
+    #[error("receiver hook aborted when {operator:?} sent {amount:?} to {to:?} from {from:?} with exit code {exit_code:?}")]
     ReceiverHook {
+        /// Whose balance is being debited
         from: ActorID,
+        /// Whose balance is being credited
         to: ActorID,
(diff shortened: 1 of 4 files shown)


fil_fungible_token/src/token/mod.rs
```diff
@@ -191,10 +194,10 @@ where
                     self.state = old_state;
                     self.flush()?;
                     Err(TokenError::ReceiverHook {
-                        from: self.msg.actor_id(),
-                        to: holder_id,
-                        sender: minter_id,
-                        value: value.clone(),
+                        from: operator_id,
+                        to: owner_id,
+                        operator: self.msg.actor_id(),
```
anorth, line 199: These are mixed up, should match the receiver hook params I think

fil_fungible_token/src/token/mod.rs
```diff
@@ -461,10 +464,10 @@ where
                     self.state = old_state;
                     self.flush()?;
                     Err(TokenError::ReceiverHook {
-                        from: owner_id,
-                        to: receiver_id,
-                        sender: spender,
-                        value: value.clone(),
+                        operator: from,
+                        to: to_id,
+                        from: operator,
```
anorth, line 469: Mixed up here too

anorth (approved):
Great

fil_fungible_token/src/token/mod.rs
anorth: (Add a test?)

PR #55, opened 2022-08-04
Basic token actor example

alexytsu:


testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -214,14 +231,12 @@ pub fn invoke(params: u32) -> u32 {
...
-                        .util
-                        .mint(&minter, &minter, &TokenAmount::from(100), &Default::default())
-                        .unwrap();
-                    token_actor.util.flush().unwrap();
-                    NO_DATA_BLOCK_ID
+                    let params = deserialize_params(params);
+                    let res = token_actor.mint(params).unwrap();
+
+                    let cid = token_actor.util.flush().unwrap();
+                    sdk::sself::set_root(&cid).unwrap();
```
abright, line 238: Without this, other calls to the actor will continue to use the previous state. I think most of the other methods that change state need to do this too?

testing/fil_token_integration/actors/basic_receiving_actor/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+fn invoke(_input: u32) -> u32 {
+    let method_num = sdk::message::method_number();
+    match_method!(method_num, {
+        "Constructor" => {
+            // this is a stateless actor so constructor does nothing
+            NO_DATA_BLOCK_ID
+        },
+        "TokensReceived" => {
+            // TokensReceived is passed a TokenReceivedParams
+            //let _params: TokenReceivedParams = deserialize_params(input);
```
abright, line 25: This seems to explode at runtime when I build the actor with it. Not sure what's wrong yet as it should be the same type we're `send`ing over when `mint()` calls the hook. Need to look into it further.

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
alexytsu: I hadn't considered this. It seems that it would be necessary to set the new root in order to handle re-entrant calls. 

However, doesn't this need to happen before the receiver hook is called (i.e. within `mint`) rather than after `mint` since the execution flow is passed to the called actor by the VM?

anorth: Yes, we'll need to set root before the receiver hook. Fun!

Alex, maybe we need to split up the mint/transfer behaviour into two parts, or take a callback that persists state, or something like that? The library can't do it internally.

abright: Mint and Transfer currently both flush state before calling the receiver hook but don't set the root at that time. Splitting mint/transfer and the receiver hook calls might be the nicest way to do things so any state persistence stuff can be handled outside of that (so token mutates state, but never stores/loads).

I've been having thoughts about making load, flush/store, and maybe a transaction thing like this into a trait for state structs, it should be generic enough that we could write the one implementation and just `impl PersistentState for MyState {}`?
The transaction option would save state (and set the root) and then revert if necessary based on the result of a callback, so we'd call the receiver hook there. I guess there might be other uses for that beyond receiver hooks too?

minting would then look something like this:
```Rust
let mint = state.token.mint(...)?;
state.call_or_revert(call_receiver_hook(mint));
```

anorth: > I've been having thoughts about making load, flush/store, and maybe a transaction thing like this into a trait

Put that onto a list for possible developer tooling things to do later. It's beyond scope for the tokens, but I'm keen to gather ideas like this.

alexytsu (commented):
LGTM for the changes on the library 

I opened this PR so can't ✅ it

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+                    let params = deserialize_params(params);
+                    let res = token_actor.decrease_allowance(params).unwrap();
+                    token_actor.util.flush().unwrap();
+                    return_ipld(&BigIntDe(res)).unwrap()
+                }
+                1691518633 => {
+                    // RevokeAllowance
+                    let params = deserialize_params(params);
+                    token_actor.revoke_allowance(params).unwrap();
+                    token_actor.util.flush().unwrap();
```
abright, line 262: Should every method call that `flush`es also `set_root` afterwards?

testing/fil_token_integration/tests/mint_tokens.rs
```diff
@@ -0,0 +1,104 @@
...
+    let return_data = ret_val.msg_receipt.return_data;
+    if return_data.is_empty() {
+        println!("return data was empty");
+    } else {
+        let mint_result: MintReturn = return_data.deserialize().unwrap();
+        println!("new total supply: {:?}", &mint_result.total_supply);
+    }
+
+    // Check balance
+    //let params = RawBytes::serialize(minter[0].1).unwrap();
```
abright, line 96: need to remove this commented code also

abright (approved):
LGTM, I think there's just a couple small things to fix up

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
alexytsu: ah, yes I think so

Issue #32, opened 2022-07-25
Decide policy for non-existent addresses


alexytsu:

anorth:
This is #31 for public key addresses. Still TBD for other addresses, but let's make it really clear in the code where we can change this behaviour.

alexytsu:
Continuing off the three functions described here, https://github.com/helix-onchain/filecoin/issues/31#issuecomment-1204597705 all addresses are handled via these helpers which implicitly gives us the behaviour
- BLS and SECP addresses have accounts initialised for them when data needs to be associated with them (i.e. storing an allowance or balance)
- Actor addresses will throw an error if an account is attempted to be initialised for them
- Any address, including Actor addresses will NOT throw an error for 'read-only' operations but instead return zero values (uninitialised accounts have no balance, no allowance etc.)

Issue #31, opened 2022-07-25
Handle resolution of non ID addresses for sends


alexytsu:
This exposes an interesting chicken and egg problem.

In the proposal, I wrote:
> An uninitialised pubkey address can receive tokens (because receiver hook will initialise it), but cannot be authorized as a spender

I was about to say you won't always have the actor ID so change to address, but I realise that we can't have the token state reflecting the new balances persisted if we don't know the receiving actor ID.

I think the required workaround will be that address resolution (in the actor code) needs to initialise actors if necessary. See `resolve_to_id_addr` in the builtin-actors `runtime/src/builtin/shared.rs`. And then this API is ok. (Defer this for a follow-up)

_Originally posted by @anorth in https://github.com/helix-onchain/filecoin/pull/28#discussion_r928323583_

alexytsu:
Pubkey addresses should probably just create the actor

Non-pubkey addresses TBD

alexytsu:
https://github.com/helix-onchain/filecoin/pull/46 addresses this with three new helpers in the `token` module

`resolve_to_id` is used when an address needs to resolve to an ID and should be created if it doesn't already exist (e.g. receiving tokens, getting allowance)

`get_id` is used when an address needs to resolve to an ID but doesn't need to be created if it can be avoided (e.g. querying the balance of an address without an ID can return 0 without creating the address)

`expect_id` is used to resolve an `Address` to an `ActorID` when we know the Address must be an id-address (e.g. the caller of a method on the actor)

anorth:
As discussed in Slack, we concluded that we should probably never return "Address uninitalised" errors to the caller, at least in relation to the owner or operator address. Instead error with insufficient balance or allowance, or succeed when the request was for zero value.

PR #57, opened 2022-08-05
Handle uninitialized accounts

alexytsu:
Addressing:
https://github.com/helix-onchain/filecoin/issues/31
https://github.com/helix-onchain/filecoin/issues/32

This implements the behaviour discussed on Slack except in the case of Actor addresses. Whereas we concluded that [Actor addresses should mirror uninitialized pubkey addresses](https://docs.google.com/spreadsheets/d/1c0A3oeB3fyO2Q3zQZubZZqUyC7AK9gGgOhgMLHtCuls/edit#gid=0) when transferring (non)-zero amounts from them, I propose that they actually forward the AddressNotInitialised error returned from the messaging layer. This inconsistency reflects a real underlying inconsistency in the behaviour of Actor-addresses. Eventually, if/when Actor-addresses are initializable, upgrading the injected messaging layer will automatically cause Actor-addresses to behave the same as other addresses without modifying the library code. 

Not having to account for Actor addresses simplifies the library code by simply initializing accounts wherever necessary and not needing a separate case to deal with zero amounts. Not treating zero amounts differently does mean however, that transfers, burns and mints of amount zero will initialise any uninitialised accounts.

A further extension to the current behaviour as a followup could potentially account for zero burns, transfers and mints as a special case and avoid unecessary account initialisations. This would not change any externally observed behaviour of the token library but can save gas/state-tree/VM-state bloat (https://github.com/helix-onchain/filecoin/pull/57#discussion_r938506584) However, on the balance, I don't think the extra complexity introduced to our Token code would be worth it.

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -345,24 +345,26 @@ where
     /// ## For all burn operations
     /// - The requested value MUST be non-negative
     /// - The requested value MUST NOT exceed the target's balance
-    ///
-    /// Upon successful burn
-    /// - The target's balance MUST decrease by the requested value
-    /// - The total_supply MUST decrease by the requested value
+    /// - If the burn operation would result in a negative balance for the owner, the burn is
+    /// discarded and this method returns an error
     ///
     /// ## Operator is the owner address
     /// If the operator is the targeted address, they are implicitly approved to burn an unlimited
     /// amount of tokens (up to their balance)
     ///
+    /// Upon successful burn
+    /// - The target's balance decreases by the requested value
+    /// - The total_supply decreases by the requested value
+    ///
     /// ## Operator burning on behalf of another address
     /// If the operator is burning on behalf of the owner the following preconditions
     /// must be met on top of the general burn conditions:
     /// - The operator MUST have an allowance not less than the requested value
-    /// In addition to the general postconditions:
-    /// - The target-operator allowance MUST decrease by the requested value
     ///
-    /// If the burn operation would result in a negative balance for the owner, the burn is
-    /// discarded and this method returns an error
+    /// Upon successful burn
+    /// - The target-operator allowance decreases by the requested value
+    /// - The target's balance decreases by the requested value
+    /// - The total_supply decreases by the requested value
     pub fn burn(
         &mut self,
         operator: &Address,
@@ -376,24 +378,45 @@ where
             )));
         }
 
+        // owner-initiated burn
+        if operator == owner {
+            let owner = self.resolve_to_id(owner)?;
+            return self.transaction(|state, bs| {
+                // attempt to burn the requested amount
+                let new_amount = state.change_balance_by(&bs, owner, &amount.clone().neg())?;
+                // decrease total_supply
+                state.change_supply_by(&amount.neg())?;
+                Ok(new_amount)
+            });
+        }
+
         // instantiate the actors if required
-        let operator = self.resolve_to_id(operator)?;
+        let operator = match self.get_id(operator) {
+            Ok(operator) => operator,
+            Err(MessagingError::AddressNotResolved(addr)) => {
+                // if not resolved, implicit allowance zero is not permitted to burn, so return an
+                // insufficient allowance error
+                return Err(TokenStateError::InsufficientAllowance {
+                    owner: *owner,
+                    operator: addr,
+                    allowance: TokenAmount::zero(),
+                    delta: amount.clone(),
+                }
+                .into());
+            }
+            Err(e) => return Err(e.into()),
+        };
+        // resolve the owner
         let owner = self.resolve_to_id(owner)?;
 
-        let new_amount = self.transaction(|state, bs| {
-            if operator != owner {
-                // attempt to use allowance and return early if not enough
-                state.attempt_use_allowance(&bs, operator, owner, amount)?;
-            }
+        self.transaction(|state, bs| {
+            state.attempt_use_allowance(&bs, operator, owner, amount)?;
             // attempt to burn the requested amount
             let new_amount = state.change_balance_by(&bs, owner, &amount.clone().neg())?;
-
             // decrease total_supply
             state.change_supply_by(&amount.neg())?;
             Ok(new_amount)
-        })?;
-
-        Ok(new_amount)
+        })
     }
 
     /// Transfers an amount from one actor to another
@@ -471,7 +494,7 @@ where
             Ok(id) => id,
             // if we cannot resolve the operator, they are forbidden to transfer
             Err(MessagingError::AddressNotResolved(operator)) => {
-                return Err(TokenError::TokenState(TokenStateError::InsufficentAllowance {
+                return Err(TokenError::TokenState(TokenStateError::InsufficientAllowance {
                     operator,
                     owner: *from,
                     allowance: TokenAmount::zero(),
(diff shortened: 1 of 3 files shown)


fil_fungible_token/src/token/mod.rs
```diff
@@ -362,8 +366,8 @@ where
             )));
         }
 
-        // owner and operator must exist to burn from
-        let operator = expect_id(operator)?;
+        // instantiate the actors if required
+        let operator = self.resolve_to_id(operator)?;
         let owner = self.resolve_to_id(owner)?;
```
alexytsu, line 371: Here in the `burn` method, an argument could be made for skipping account initialization when handling zero-amounts. 

If the both `operator` and `owner` are uninitialised but `owner` is an Actor-address, the `operator` address will be initialised whilst the `burn` method fails. In some implementations, our consumers may catch this error and not abort the entire VM-level transaction. This may lead to unexpected state being persisted on-chain.

anorth:
Ok, I am aligned with this treatment of actor addresses, at least for now. But maybe we can change something in the VM for them: see https://github.com/filecoin-project/FIPs/discussions/379#discussioncomment-3345480

anorth (approved):
Approving this progress, but thinking we should change the behaviour for zero-allowance operators.

alexytsu:
@anorth ready for re-review with https://github.com/helix-onchain/filecoin/pull/62 adding the table tests

fil_fungible_token/src/token/mod.rs
```diff
@@ -362,24 +378,45 @@ where
             )));
         }
 
-        // owner and operator must exist to burn from
-        let operator = expect_id(operator)?;
-        let owner = self.resolve_to_id(owner)?;
+        // owner-initiated burn
+        if operator == owner {
```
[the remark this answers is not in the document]
anorth: Oh, but if owner fails to resolve, there can be no balance to burn (but we can still allow zero).

Note that there's no receiver hook here, but I support matching logic as if there were a sender hook anyway.

fil_fungible_token/src/token/mod.rs
```diff
@@ -362,24 +378,45 @@ where
...
+                state.change_supply_by(&amount.neg())?;
+                Ok(new_amount)
+            });
+        }
 
-        let new_amount = self.transaction(|state, bs| {
-            if operator != owner {
-                // attempt to use allowance and return early if not enough
-                state.attempt_use_allowance(&bs, operator, owner, amount)?;
+        // instantiate the actors if required
```
anorth, line 393: This comment isn't right.

fil_fungible_token/src/token/mod.rs
```diff
@@ -362,24 +378,45 @@ where
...
+                }
+                .into());
             }
+            Err(e) => return Err(e.into()),
+        };
+        // resolve the owner
+        let owner = self.resolve_to_id(owner)?;
+
+        self.transaction(|state, bs| {
+            state.attempt_use_allowance(&bs, operator, owner, amount)?;
```
anorth, line 413: `attempt_use_allowance` needs to be updated to fail if current allowance is zero.

fil_fungible_token/src/token/mod.rs
```diff
@@ -362,24 +378,45 @@ where
...
+                    operator: addr,
+                    allowance: TokenAmount::zero(),
+                    delta: amount.clone(),
+                }
+                .into());
             }
+            Err(e) => return Err(e.into()),
+        };
+        // resolve the owner
+        let owner = self.resolve_to_id(owner)?;
```
anorth, line 410: From our sheet, the only case where we should init is `owner == operator`.

fil_fungible_token/src/token/mod.rs
```diff
@@ -389,23 +426,26 @@ where
...
+    /// interface specified for FRC-XXX token receiver. If the receiving hook aborts, when called,
+    /// the transfer is discarded and this method returns an error
     ///
     /// ## Operator equals owner address
     /// If the operator is the owner address, they are implicitly approved to transfer an unlimited
     /// amount of tokens (up to their balance)
     ///
+    /// Upon successful transfer:
+    /// - The senders's balance decreases by the requested value
+    /// - The receiver's balance increases by the requested value
```
anorth, line 438: Duplicated below

fil_fungible_token/src/token/mod.rs
```diff
@@ -420,85 +460,89 @@ where
...
+                    owner: *from,
+                    allowance: TokenAmount::zero(),
+                    delta: amount.clone(),
+                }));
+            }
+            Err(e) => return Err(e.into()),
+        };
+
+        // resolve the participating actors
         let from = self.resolve_to_id(from)?;
```
anorth, line 508: Just get_id here

fil_fungible_token/src/token/mod.rs
```diff
@@ -1162,25 +1374,158 @@ mod test {
...
+                assert_eq!(delta, burn_amount);
+            }
+            e => panic!("unexpected error {:?}", e),
+        };
+        // balances unchanged
+        assert_eq!(token.total_supply(), TokenAmount::from(400_000));
+        assert_eq!(token.balance_of(TREASURY).unwrap(), TokenAmount::from(400_000));
+        assert_eq!(token.allowance(TREASURY, secp_address).unwrap(), TokenAmount::zero());
+
+        // but can burn zero
```
anorth, line 1459: This should fail (for consistency with denying transfer of zero with zero allowance)

anorth (changes requested):

fil_fungible_token/src/token/mod.rs
```diff
@@ -1140,6 +1226,132 @@ mod test {
...
+    fn it_allows_delegated_transfer_by_resolvable_pubkey() {
+        let mut token = new_token();
+
+        // mint 100 for owner
+        token.mint(TOKEN_ACTOR, ALICE, &TokenAmount::from(100), &Default::default()).unwrap();
+
+        let initialised_address = &secp_address();
+        token.msg.initialize_account(initialised_address).unwrap();
+
+        // an initialised pubkey can transfer zero out of Alice balance
```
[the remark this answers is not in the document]
alexytsu: yup, thankfully the table tests in https://github.com/helix-onchain/filecoin/pull/62 picked it up and was fixed but I've brought the change into this PR now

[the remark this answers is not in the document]
alexytsu: I can merge that PR into this one or wait until you've reviewed that diff separately

fil_fungible_token/src/token/mod.rs
alexytsu: as above, this is addressed in #62

fil_fungible_token/src/token/mod.rs
```diff
@@ -115,22 +115,81 @@ where
...
     }
 
-    /// Attempts to resolve an address to an ID address, returning MessagingError::AddressNotInitialized
+    /// Attempts to resolve an address to an ActorID, returning MessagingError::AddressNotResolved
     /// if it wasn't found
     fn get_id(&self, address: &Address) -> MessagingResult<ActorID> {
         self.msg.resolve_id(address)
     }
+
+    /// Attempts to compare two addresses, seeing if they would resolve to the same Actor without
```
anorth, line 133: Style Nit: Consider moving helpers toward the bottom of the file (a new impl block?) so that the code reads more in a top-down fashion, when read in order.

fil_fungible_token/src/token/mod.rs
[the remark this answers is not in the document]
anorth: Thanks. The code is right now but the comment still old.

anorth (approved):
Well, this ended up adding quite a bit of noise, but I think the behaviour is nice and consistent. Maybe there's some scope to factor out some of the repeated address-resolution blocks later.

Issue #56, opened 2022-08-05
allow Token to wrap an already loaded state


alexytsu:

alexytsu:
This is now the "default" as of #60

PR #59, opened 2022-08-09
have token handle hold a refcell to state

alexytsu:
As an alternative to #60

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -1,5 +1,5 @@
-mod state;
-mod types;
+pub mod state;
+pub mod types;
 
 use self::state::{StateError, TokenState};
 use crate::receiver::types::TokenReceivedParams;
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -73,11 +73,16 @@ where
     BS: Blockstore,
     MSG: Messaging,
 {
+    /// Take an existing loaded state tree and create a token handle
+    pub fn new(bs: BS, msg: MSG, token_state: TokenState) -> Self {
+        Self { bs, msg, state: RefCell::new(token_state) }
+    }
+
     /// Creates a new token instance using the given blockstore and creates a new empty state tree
     ///
     /// Returns a Token handle that can be used to interact with the token state tree and the Cid
     /// of the state tree root
-    pub fn new(bs: BS, msg: MSG) -> Result<(Self, Cid)> {
+    pub fn create(bs: BS, msg: MSG) -> Result<(Self, Cid)> {
         let init_state = TokenState::new(&bs)?;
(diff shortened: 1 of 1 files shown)


alexytsu:
https://github.com/helix-onchain/filecoin/issues/56

fil_fungible_token/src/token/mod.rs
```diff
@@ -535,6 +548,29 @@ mod test {
...
+        // wrap the token state, moving it into a TokenHandle
+        let mut token = Token::wrap(
+            &bs,
+            FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 6),
+            actor_state.token_state,
+        );
+
+        token.mint(TOKEN_ACTOR, TREASURY, &TokenAmount::from(1), &Default::default()).unwrap();
+        // gets a read-only state
+        let state = token.borrow_state();
```
anorth, line 570: Hmm, but what about `actor_state.token_state`. Did `RefCell::new` copy the state (it implements Clone, but not Copy...)? `Wrap` wouldn't be the right word to describe what's happening if so, and I see decent potential for error here inspecting the old state. Did the old one get moved so it can't be read any more?

Could we explore more what's the deal with the two states now? Included commented-out lines for things that don't compile, to explain that.

We don't need to eliminate clones at all costs, but would like a justification for them.

alexytsu: `RefCell::new` will take ownership of the state so it's da move (memcpy) rather than a clone. 

`actor_state.token_state` is not valid after line 565 anymore.

anorth: Ok, even memcpy is a bit undesirable, but at least it's safe.

anorth: Can a different field in actor_state be accessed? If the partial move invalidates the whole containing struct, it will be awkward for uses that have embedded token state.

fil_fungible_token/src/token/mod.rs
```diff
@@ -535,12 +548,38 @@ mod test {
...
+        assert_eq!(state.supply, TokenAmount::from(1));
+
+        // actor_state.token_state is moved when `wrap` is called so cannot be accessed anymore
+        // assert_eq!(actor_state.token_state.supply, TokenAmount::zero());
```
anorth, line 574: But did it write back to actor_state or not? If I now want to persist that state, I need to be able to access actor_state to serialize it.

alexytsu:
I don't think this is acceptable because the partial move of token state out of the actor's state tree will always invalidate the entire actor state tree

fil_fungible_token/src/token/mod.rs
alexytsu: Yeh, the entire actor_state becomes invalid after the partial move so this is pretty awkward compared to the &mut impl

PR #60, opened 2022-08-09
construct token via mutable ref to token state

alexytsu:
As previously discussed in Slack

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -131,43 +131,19 @@ where
 
         // Update state so re-entrant calls see the changes
         self.flush()?;
+
         // Call receiver hook
-        let hook_params = TokenReceivedParams {
-            data: data.clone(),
-            from: self.msg.actor_id(),
-            to: owner_id,
-            operator: operator_id,
-            amount: amount.clone(),
-        };
-        match self.msg.send(
+        self.call_receiver_hook_or_revert(
             initial_owner,
-            RECEIVER_HOOK_METHOD_NUM,
-            &RawBytes::serialize(hook_params)?,
-            &TokenAmount::zero(),
-        ) {
-            Ok(receipt) => {
-                // hook returned true, so we can continue
-                if receipt.exit_code.is_success() {
-                    Ok(())
-                } else {
-                    *self.state = old_state;
-                    self.flush()?;
-                    Err(TokenError::ReceiverHook {
-                        operator: operator_id,
-                        from: self.msg.actor_id(),
-                        to: owner_id,
-                        amount: amount.clone(),
-                        exit_code: receipt.exit_code,
-                    })
-                }
-            }
-            Err(e) => {
-                // error calling receiver hook, revert state
-                *self.state = old_state;
-                self.flush()?;
-                Err(e.into())
-            }
-        }
+            TokenReceivedParams {
+                data: data.clone(),
+                from: self.msg.actor_id(),
+                to: owner_id,
+                operator: operator_id,
+                amount: amount.clone(),
+            },
+            old_state,
+        )
     }
 
     /// Gets the total number of tokens in existence
@@ -646,13 +622,6 @@ mod test {
     use crate::token::Token;
     use crate::token::TokenError;
 
-    fn new_token(
-        bs: MemoryBlockstore,
-        state: &mut TokenState,
-    ) -> Token<MemoryBlockstore, FakeMessenger> {
-        Token::wrap(bs, FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 6), 1, state)
-    }
-
     /// Returns a static secp256k1 address
     fn secp_address() -> Address {
         let key = vec![0; 65];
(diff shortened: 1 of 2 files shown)


alexytsu:
https://github.com/helix-onchain/filecoin/issues/56

fil_fungible_token/src/token/mod.rs
```diff
@@ -76,25 +76,21 @@ where
...
-
-    /// For an already initialised state tree, loads the state tree from the blockstore and returns
-    /// a Token handle to interact with it
-    pub fn load(bs: BS, msg: MSG, state_cid: Cid) -> Result<Self> {
-        let state = TokenState::load(&bs, &state_cid)?;
-        Ok(Self { bs, msg, state })
+    pub fn new(bs: BS, msg: MSG, state: &'st mut TokenState) -> Result<Self> {
+        let token = Self { bs, msg, state };
+        Ok(token)
     }
```
anorth, line 82: If we go this way, I'd expect to add an `init` or similar method that creates new state as the old `new` did, and returns both. And similarly for `load` probably.

fil_fungible_token/src/token/mod.rs
```diff
@@ -524,23 +521,50 @@ mod test {
...
+            &bs,
+            FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 6),
+            &mut actor_state.token_state,
+        )
+        .unwrap();
+
+        token.mint(TOKEN_ACTOR, TREASURY, &TokenAmount::from(1), &Default::default()).unwrap();
+
+        let state = token.state();
+        // gets a read-only state
```
anorth, line 557: Can we access `actor_state` or is that now disallowed due to the mutable reference? (Fine if not, but please demonstrate). If we can access it, can we show the change is visible there too, ie. there's only one copy?

fil_fungible_token/src/token/mod.rs
alexytsu: I think this is not possible anymore the way it's currently structured (@abright maybe you can see a way perhaps with heap allocation via a `Box`). The awkwardness of _always_ having to create the `TokenState` externally and pass it in to a `Token` is the primary reason I turned to a `RefCell` impl

If a helper function instantiates the `TokenState` that token state is dropped at the end of whatever helper function creates it. Therefore the lifetime of any reference to it becomes invalid and can't be used in the `Token` that we actually want returned.

fil_fungible_token/src/token/mod.rs
alexytsu: `actor_state.token_state` can still be accessed (see line 560 now) but doing so invalidates the `token` created on line 547. 

not quite what I expected/wanted on first consideration but makes sense that it works like this.

fil_fungible_token/src/token/mod.rs
anorth: But it won't be dropped if it's returned....

anorth: I guess you won't be able to return the Token then. Ok fine, consider it just a convenience for creating a state.

alexytsu: I'm not sure if this is what you are thinking of https://github.com/helix-onchain/filecoin/pull/60/commits/3b330574f72d31fa5991f7c02fd6ec2065411b54

fil_fungible_token/src/token/mod.rs
```diff
@@ -76,25 +76,33 @@ where
     ///
     /// Returns a Token handle that can be used to interact with the token state tree and the Cid
     /// of the state tree root
-    pub fn new(bs: BS, msg: MSG) -> Result<(Self, Cid)> {
-        let init_state = TokenState::new(&bs)?;
-        let cid = init_state.save(&bs)?;
-        let token = Self { bs, msg, state: init_state };
-        Ok((token, cid))
+    pub fn new(bs: BS, msg: MSG, state: &'st mut TokenState) -> Result<Self> {
```
anorth, line 79: Don't need result any more

fil_fungible_token/src/token/mod.rs
```diff
@@ -524,23 +533,58 @@ mod test {
...
+        token.mint(TOKEN_ACTOR, TREASURY, &TokenAmount::from(1), &Default::default()).unwrap();
+
+        let state = token.state();
+        // gets a read-only state
+        assert_eq!(state.supply, TokenAmount::from(1));
+        // can get a token_state here but doing so borrows the value making the mutable borrow on line 550 invalid
+        assert_eq!(actor_state.token_state.supply, TokenAmount::from(1));
+
+        // therefore, after the above line 560, can no longer use the token handle to read OR mutate state
+        // any single one of these lines now causes a compiler error
```
anorth, line 575: This seems acceptable

anorth (approved):

alexytsu:
I _think_ this change makes the current "wrap" unnecessary but perhaps a higher level transaction API on the `Token` handle itself is now needed.

fil_fungible_token/src/token/mod.rs
```diff
@@ -40,64 +40,47 @@ where
     /// Minimal interface to call methods on other actors (i.e. receiver hooks)
     msg: MSG,
     /// In-memory cache of the state tree
```
anorth, line 42: "Reference to token state that will be inspected/mutated"

fil_fungible_token/src/token/mod.rs
```diff
@@ -676,12 +658,46 @@ mod test {
...
+    #[test]
+    fn it_wraps_a_previously_loaded_state_tree() {
+        struct ActorState {
+            token_state: TokenState,
+        }
+
+        // simulate the token state being a node in a larger state tree
+        let bs = MemoryBlockstore::default();
+        let mut actor_state =
+            ActorState { token_state: Token::<_, FakeMessenger>::create_state(&bs).unwrap() };
```
anorth, line 670: I guess this "convenience" method isn't that convenient due to generics.

anorth (approved):

PR #64, opened 2022-08-10
Add granularity enforcement to token library

anorth:
I was about to implement this in actor code, but realised the library should do it once for everyone.

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -1,10 +1,4 @@
-mod state;
-mod types;
-
-use self::state::{StateError, TokenState};
-use crate::receiver::types::TokenReceivedParams;
-use crate::runtime::messaging::{Messaging, MessagingError};
-use crate::runtime::messaging::{Result as MessagingResult, RECEIVER_HOOK_METHOD_NUM};
+use std::ops::{Neg, Rem};
 
 use cid::Cid;
 use fvm_ipld_blockstore::Blockstore;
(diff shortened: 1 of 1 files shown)


fil_fungible_token/src/token/mod.rs
```diff
@@ -17,15 +11,30 @@ use fvm_shared::error::ExitCode;
...
 
+use crate::receiver::types::TokenReceivedParams;
+use crate::runtime::messaging::{Messaging, MessagingError};
+use crate::runtime::messaging::{Result as MessagingResult, RECEIVER_HOOK_METHOD_NUM};
+use crate::token::TokenError::InvalidGranularity;
+
+use self::state::{StateError, TokenState};
+
+mod state;
+mod types;
```
anorth, line 24: My IDE (JetBrains rust plugin) reorganised this. What are you using?

alexytsu: I think vscode calls out to rustfmt which is what CI uses as a checker

with [group imports](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#group_imports) set to StdExternalCrate the output is close except that it groups self and crate imports together. I'll just add it to our rusfmt.toml if you are able to configure your plugin to use rustfmt.

fil_fungible_token/src/token/mod.rs
```diff
@@ -355,16 +352,10 @@ where
...
-            return Err(TokenError::InvalidNegative(format!(
-                "burn amount {} cannot be negative",
-                amount
-            )));
-        }
-
         // owner and operator must exist to burn from
         let operator = expect_id(operator)?;
         let owner = self.resolve_to_id(owner)?;
+        let amount = validate_amount(amount, "burn", self.granularity)?;
```
alexytsu, line 358: is this check purposely after the `resolve_to_id` calls?

in my mind, it's lighter to have the validate_amount check fail first always

alexytsu (approved):
approving but wondering if there's something i haven't considered re: ordering of amount validation vs account validation

fil_fungible_token/src/token/mod.rs
anorth: Let's try it, we can go back to default if it's not great.

fil_fungible_token/src/token/mod.rs
anorth: I moved it to have the param validation match the param order, but you make a good point and I'll move it back.

Issue #58, opened 2022-08-07
Add standard mapping of token lib errors to actor exit codes


anorth:
In the built-in actors, callers currently do something like
```
                token
                    .mint(operator, to, &params.amount, RawBytes::default())
                    .exit_code(ExitCode::USR_UNSPECIFIED) // FIXME translate error to exit code
```

(where `exit_code` is method on `Result<T, std::error::Error>` that adapts to an `ActorError` (built-in actors type that carries exit code and results in abort). 

Without doing anything too specific to the built-in actors runtime, add a standard mapping of errors to exit codes, so that all token implementers will use consistent code by default. A free function doing the mapping could be sufficient.

PR #66, opened 2022-08-10
map TokenError to ExitCodes

alexytsu:
https://github.com/helix-onchain/filecoin/issues/58

I don't know if I'm going down the right path. I'm not quite sure how well this fits in to what you need but this could be used:

```
token
    .mint(operator, to, &params.amount, RawBytes::default())
    .map_err(ExitCode::from)
```

Which results in a `Result<T, ExitCode>`

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -13,13 +13,12 @@ use num_traits::Signed;
 use num_traits::Zero;
 use thiserror::Error;
 
+use self::state::StateInvariantError;
+use self::state::{StateError as TokenStateError, TokenState};
 use crate::receiver::types::TokenReceivedParams;
 use crate::runtime::messaging::{Messaging, MessagingError};
 use crate::runtime::messaging::{Result as MessagingResult, RECEIVER_HOOK_METHOD_NUM};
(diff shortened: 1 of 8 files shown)


anorth:
It's likely that callers will map into a struct containing an ExitCode (and context message) rather than bare code, but this is still good, they can use `into()`.

fil_fungible_token/src/token/mod.rs
```diff
@@ -49,6 +50,53 @@ pub enum TokenError {
...
+                }
+                StateError::MissingState(_) => ExitCode::USR_ILLEGAL_STATE,
+                StateError::NegativeBalance { balance: _, delta: _, owner: _ }
+                | StateError::InsufficentAllowance {
+                    owner: _,
+                    operator: _,
+                    allowance: _,
+                    delta: _,
+                } => ExitCode::USR_FORBIDDEN,
+                StateError::NegativeTotalSupply { supply: _, delta: _ } => ExitCode::USR_FORBIDDEN,
```
anorth, line 74: Where does this come from? It seems like an internal invariant failure, in which case ILLEGAL_STATE or ASSERTION_FAILED

anorth (approved):

Issue #63, opened 2022-08-10
Implement a state invariant check that verifies all internal invariants


anorth:
See the built-in actors and https://github.com/filecoin-project/builtin-actors/pull/513#discussion_r941743698 as inspiration.

Add a method to TokenState that checks all invariants. Call this at the end of every unit test, and expose it for use externally.

alexytsu:
https://github.com/helix-onchain/filecoin/issues/63

PR #67, opened 2022-08-10
add invariant checks for TokenState

alexytsu:
Adds some invariant checking thats suitable to be called during unit/integration testing
https://github.com/helix-onchain/filecoin/issues/63

### DIFF
--- a/fil_fungible_token/src/token/state.rs
+++ b/fil_fungible_token/src/token/state.rs
@@ -1,3 +1,6 @@
+use std::ops::Neg;
+
+use anyhow::bail;
 use cid::multihash::Code;
 use cid::Cid;
 use fvm_ipld_blockstore::Block;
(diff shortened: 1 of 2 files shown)


fil_fungible_token/src/token/state.rs
```diff
@@ -41,6 +44,32 @@ pub enum StateError {
...
+        "stored a zero allowance which should have been removed between {owner:?} and {operator:?}"
+    )]
+    ExplicitZeroAllowance { owner: ActorID, operator: ActorID },
+    #[error("stored an allowance map for {0} though they have specified no allowances")]
+    ExplicitEmptyAllowance(ActorID),
+    #[error("stored an allowance for self {account:?} for {allowance:?}")]
+    ExplicitSelfAllowance { account: ActorID, allowance: TokenAmount },
+    #[error("underlying state error {0}")]
+    State(#[from] StateError),
+}
```
anorth, line 71: Thanks. Since these are not really used in production, and any error is bad, this level of structure is probably overkill. But now that its here, great.

anorth (approved):
Great

PR #69, opened 2022-08-10
Add with() method for token ops wrapping state

anorth:
This unblocks actors being able to work with a token state embedded in another struct. It's otherwise impossible to construct a Token with that state. This isn't the final form, but maybe the right API.

I considered a `new` that clone the state, but rejected due to the potential for failing to write the changes back.


alexytsu (approved):
LGTM, the `with` looks neat and as you've noted we can refactor for more efficiency

PR #72, opened 2022-08-15
Split CI into more, smaller jobs

abright:
This makes separate jobs for the build, test coverage and actor test stages so that they can run in parallel


abright:
This looks to take about half the time of what's currently in `main`

alexytsu:
I imagined this might be cleaner if parallelization happened at the job level rather than the workflow level so that there were simply three workflows. 
1. lint check on push
2. build + coverage/actor test on PR
3. build + coverage/actor test on main

This does mean that the time for 2. and 3. increases slightly (test may happen sequentially after build). However I think the coverage/actor test may gain some speed up in compilation stage from having the build step run before it. I think keeping them on the same workflow means they execute sequentially on the same runner.

abright:
I think each of the three steps ends up building everything (or almost everything) each time because the each run a different combination of target and RUSTFLAGS. So there might not be much (or any) sharing going on either way.

Will try it as separate jobs in one workflow though

abright:
Seems to take around the same time for this approach, but it's cleaner code so let's go with that

alexytsu (approved):
LGTM

PR #75, opened 2022-08-18
Update the token-actor interface

alexytsu:
https://github.com/filecoin-project/FIPs/discussions/407 should be sync'd with the new interface/types defined here


testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -46,23 +47,50 @@ impl FrcXXXToken<RuntimeError> for BasicToken<'_> {
         Ok(self.util.balance_of(&params)?)
     }
 
+    fn transfer(&mut self, params: TransferParams) -> Result<TransferReturn, RuntimeError> {
+        let spender = caller_address();
+        self.util.transfer(&spender, &spender, &params.to, &params.amount, &params.data)?;
+
+        // FIXME: populate with real values
```
alexytsu, line 54: The library function here would need to return the new balances and remaining allowance in order to support this return type well

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -47,23 +50,46 @@ impl FrcXXXToken<RuntimeError> for BasicToken<'_> {
...
+        Ok(TransferReturn { from_balance: TokenAmount::zero(), to_balance: TokenAmount::zero() })
```
alexytsu, line 58: might be more efficient if Token::transfer returned the relevant info, else we can make separate calls to `Token::balance_of` and `Token::allowance` etc.

fil_fungible_token/src/receiver/types.rs
```diff
@@ -6,26 +6,26 @@ use fvm_shared::ActorID;
...
+    /// Address of the operator that initiated the transfer/mint
     pub operator: ActorID,
     /// The account that the tokens are being pulled from (the token actor address itself for mint)
     pub from: ActorID,
     /// The account that the tokens are being sent to (the receiver address)
     pub to: ActorID,
     #[serde(with = "bigint_ser")]
     pub amount: TokenAmount,
+    /// Data specified by the operator during transfer/mint
     pub data: RawBytes,
```
alexytsu, line 29: was there a need for a separate byte array to carry data from the token-actor itself?

anorth: It's a good question, ERC-777 has one. Can you ask in https://github.com/filecoin-project/FIPs/discussions/407?

fil_fungible_token/src/token/types.rs
```diff
@@ -20,108 +20,148 @@ pub type Result<T, E> = std::result::Result<T, ActorError<E>>;
...
 
     /// Returns the ticker symbol of the token
+    ///
+    /// Must not be empty. Should be a short uppercase string
     fn symbol(&self) -> String;
 
     /// Returns the total amount of the token in existence
+    ///
+    /// Must be non-negative. The total supply must equal the balances of all addresses. The total
+    /// supply must equal the sum of all minted tokens less the sum of all burnt tokens.ß
```
anorth, line 39:
```suggestion
    /// supply should equal the sum of all minted tokens less the sum of all burnt tokens.
```
This is relaxed in the FIP too. There's really very little we can constrain outside of the method signatures.

fil_fungible_token/src/token/types.rs
```diff
@@ -20,108 +20,148 @@ pub type Result<T, E> = std::result::Result<T, ActorError<E>>;
...
+    fn transfer_from(&mut self, params: TransferFromParams) -> Result<TransferFromReturn, E>;
 
-    /// Set the allowance a operator has on the owner's account to zero
+    /// Atomically increases the approved allowance that a operator can transfer/burn from the
+    /// caller's balance
+    ///
+    /// The increase must be non-negative. Returns the new total allowance approved for that
+    /// owner-operator pair.
+    fn increase_allowance(&mut self, params: IncreaseAllowanceParams)
+        -> Result<AllowanceReturn, E>;
```
anorth, line 66: I think this should return just the TokenAmount. Same for decrease_allowance.

fil_fungible_token/src/token/types.rs
```diff
@@ -20,108 +20,148 @@ pub type Result<T, E> = std::result::Result<T, ActorError<E>>;
...
 
-    /// Get the allowance between two addresses
+    /// Returns the allowance approved for an operator on a spender's balance
     ///
-    /// The operator can burn or transfer the allowance amount out of the owner's address. If the
-    /// address of the owner cannot be resolved, this method returns an error. If the owner can be
-    /// resolved, but the operator address is not registered with an allowance, an implicit allowance
-    /// of 0 is returned
+    /// The operator can burn or transfer the allowance amount out of the owner's address.
     fn allowance(&mut self, params: GetAllowanceParams) -> Result<AllowanceReturn, E>;
```
anorth, line 83: And this one just a TokenAmount too, I reckon

fil_fungible_token/src/token/types.rs
```diff
@@ -140,50 +187,47 @@ pub struct AllowanceReturn {
...
 #[derive(Serialize_tuple, Deserialize_tuple)]
-pub struct TransferParams {
-    pub from: Address,
-    pub to: Address,
+pub struct BurnReturn {
+    /// New balance in the account after the successful burn
     #[serde(with = "bigint_ser")]
-    pub amount: TokenAmount,
-    pub data: RawBytes,
+    pub remaining_balance: TokenAmount,
```
anorth, line 218: Just `balance`, I reckon

fil_fungible_token/src/token/types.rs
```diff
@@ -140,50 +187,47 @@ pub struct AllowanceReturn {
...
 #[derive(Serialize_tuple, Deserialize_tuple)]
-pub struct TransferReturn {
-    pub from: Address,
-    pub to: Address,
-    pub by: Address,
+pub struct BurnFromReturn {
+    /// New balance in the account after the successful burn
     #[serde(with = "bigint_ser")]
-    pub amount: TokenAmount,
+    pub remaining_balance: TokenAmount,
```
anorth, line 226: Ditto `balance`.

fil_fungible_token/src/token/types.rs
```diff
@@ -140,50 +187,47 @@ pub struct AllowanceReturn {
...
+    /// New remaining allowance between the owner and operator (caller)
+    #[serde(with = "bigint_ser")]
+    pub remaining_allowance: TokenAmount,
```
anorth, line 229: I'm ambivalent about the `remaining_` prefix on this one (so long as consistent with TransferFrom)

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
anorth: Agreed. Simple but maybe error-prone would be to return an anonymous tuple. Another alternative is to return the Return types defined earlier, but that inverts the expected dependency hierarchy. Another option is separate return structs, but they'll be a bit duplicative.

anorth (commented):
Great improvement, thanks.

fil_fungible_token/src/receiver/types.rs
alexytsu: https://github.com/filecoin-project/FIPs/discussions/407#discussioncomment-3443327

fil_fungible_token/src/receiver/types.rs
```diff
@@ -5,28 +5,30 @@ use fvm_shared::econ::TokenAmount;
...
-    fn token_received(params: TokenReceivedParams);
+    fn tokens_received(params: TokensReceivedParams);
 }
 
 #[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
-pub struct TokenReceivedParams {
-    /// Address of the operator that initiated the transfer
+pub struct TokensReceivedParams {
+    /// Address of the operator that initiated the transfer/mint
     pub operator: ActorID,
```
anorth, line 21: Could we align on a consistent ordering of fields here and in all the Params/Return types. I suggest: From, To, Operator, Amount, OperatorData, TokenData

fil_fungible_token/src/token/types.rs
```diff
@@ -18,172 +18,204 @@ pub enum ActorError<Err> {
...
-/// //        2 => token.name(),
-/// //        _ => abort!()
-/// //        // etc.
-/// //    }
-/// //}
-/// ```
-pub trait FrcXXXToken<E> {
+/// Token authors must implement this trait and link the methods to standard dispatch numbers (as
+/// defined by [FRC-0042](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0042.md)).
+pub trait Frc46Token<E> {
```
anorth, line 25: I don't think Rust has opinions about acronyms, like Go does. But I do! Could we capitalise the FRC acronym please?

fil_fungible_token/src/token/mod.rs
```diff
@@ -606,6 +629,62 @@ fn validate_amount<'a>(
...
+    type Error = TokenError;
+
+    fn try_from(value: BurnResult) -> Result<Self> {
+        match value {
+            BurnResult::BurnFrom(ret) => Ok(ret),
+            BurnResult::Burn(_) => Err(TokenError::BurnReturn { result: value }),
+        }
+    }
+}
+
```
alexytsu, line 687: An alternative to this might be to have separate library level methods for `transfer` and `transfer_from` which return different types that match the FRC interface

anorth (approved):

fil_fungible_token/src/token/mod.rs
```diff
@@ -112,7 +116,8 @@ where
         operator: &Address,
         initial_owner: &Address,
         amount: &TokenAmount,
-        data: &RawBytes,
+        operator_data: RawBytes,
+        token_data: RawBytes,
```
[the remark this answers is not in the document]
alexytsu: It was intentional but I'm not sure if my reasoning is too pedantic or even idiomatic with rust conventions:

The addresses and token amounts are borrowed because the method doesn't really "operate on" them and only reads their values. 

I think of the data as consumed by the method, which needs to "send" it. My intuition for this comes from the fact that constructing the `TokensReceivedParams` now can simply consume the moved value rather than cloning from the reference which seems neater and more idiomatic with my understanding of borrowing vs ownership. 

Counterpoint is that we often clone from references to `TokenAmount` within this library. However, I tend to think of the TokenAmount as a primitive and clone as necessary wherever a primitive would have been copied.

fil_fungible_token/src/token/types.rs
```diff
@@ -18,172 +18,205 @@ pub enum ActorError<Err> {
 pub type Result<T, E> = std::result::Result<T, ActorError<E>>;
```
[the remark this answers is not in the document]
alexytsu: I'll do some thinking and experimenting around this but deferred for a followup

fil_fungible_token/src/token/mod.rs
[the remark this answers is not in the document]
anorth: Makes sense, thanks.

PR #76, opened 2022-08-23
Split receiver hook calls from minting and transfer - part 1

abright:
Splits the mint and transfer operations and receiver hook calls, so we can handle saving state and making the receiver hook call up at the actor level. This part 1 PR does the split but could use some API improvements like a return type to wrap up the receiver hook stuff as suggested in https://github.com/helix-onchain/filecoin/pull/76#pullrequestreview-1082792669 below.

There's also a behaviour change here: we'll ultimately abort on failure results and won't revert state ourselves

NOTE: it breaks a few of the existing unit tests that expect the state to revert on failure/rejection, they're ignored for now and can be addressed in part 2.

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -643,17 +643,12 @@ where
         token_receiver: &Address,
         params: TokensReceivedParams,
     ) -> Result<()> {
-        let receipt = match self.msg.send(
+        let receipt = self.msg.send(
             token_receiver,
             RECEIVER_HOOK_METHOD_NUM,
             &RawBytes::serialize(&params)?,
             &TokenAmount::zero(),
-        ) {
-            Ok(receipt) => receipt,
-            Err(e) => {
-                return Err(e.into());
-            }
-        };
+        )?;
 
         match receipt.exit_code {
             ExitCode::OK => Ok(()),
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -374,10 +374,8 @@ where
         amount: &TokenAmount,
         operator_data: RawBytes,
         token_data: RawBytes,
-    ) -> Result<TransferReturn> {
+    ) -> Result<(TokensReceivedParams, TransferReturn)> {
         let amount = validate_amount(amount, "transfer", self.granularity)?;
-
-        let old_state = self.state.clone();
 
         // owner-initiated transfer
         let from = self.resolve_or_init(from)?;
@@ -403,10 +401,7 @@ where
             }
         })?;
 
-        // call receiver hook
-        self.flush()?;
-        self.call_receiver_hook_or_revert(
-            to,
+        Ok((
             TokensReceivedParams {
                 operator: from,
                 from,
@@ -415,10 +410,8 @@ where
                 operator_data,
                 token_data,
             },
-            old_state,
-        )?;
-
-        Ok(res)
+            res,
+        ))
     }
 
     /// Transfers an amount from one address to another
@@ -442,13 +435,11 @@ where
         amount: &TokenAmount,
         operator_data: RawBytes,
         token_data: RawBytes,
-    ) -> Result<TransferFromReturn> {
+    ) -> Result<(TokensReceivedParams, TransferFromReturn)> {
         let amount = validate_amount(amount, "transfer", self.granularity)?;
         if self.same_address(operator, from) {
             return Err(TokenError::InvalidOperator(*operator));
         }
-
-        let old_state = self.state.clone();
 
         // operator-initiated transfer must have a resolvable operator
         let operator_id = match self.get_id(operator) {
@@ -510,10 +501,7 @@ where
             }
         })?;
 
-        // flush state as receiver hook needs to see new balances
-        self.flush()?;
-        self.call_receiver_hook_or_revert(
-            to,
+        Ok((
             TokensReceivedParams {
                 operator: operator_id,
                 from,
@@ -522,10 +510,8 @@ where
                 operator_data,
                 token_data,
             },
-            old_state,
-        )?;
-
-        Ok(ret)
+            ret,
+        ))
     }
 }
 
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -118,27 +118,40 @@ where
         amount: &TokenAmount,
         operator_data: RawBytes,
         token_data: RawBytes,
-    ) -> Result<()> {
+    //) -> Result<()> {
+        ) -> Result<TokensReceivedParams> {
         let amount = validate_amount(amount, "mint", self.granularity)?;
         // init the operator account so that its actor ID can be referenced in the receiver hook
         let operator_id = self.resolve_or_init(operator)?;
         // init the owner account as allowance and balance checks are not performed for minting
         let owner_id = self.resolve_or_init(initial_owner)?;
 
-        let old_state = self.state.clone();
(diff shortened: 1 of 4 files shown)


fil_fungible_token/src/token/mod.rs
```diff
@@ -113,27 +113,39 @@ where
...
         let operator_id = self.resolve_or_init(operator)?;
         // init the owner account as allowance and balance checks are not performed for minting
         let owner_id = self.resolve_or_init(initial_owner)?;
 
-        let old_state = self.state.clone();
+        // state save/revert will be managed outside of this function
+        //let old_state = self.state.clone();
 
         // Increase the balance of the actor and increase total supply
+        // todo: probably don't need transaction thing here, the entire minting+receiver hook operation belongs in one
```
anorth, line 128: I think it is still a useful property that this token object maintains state that makes sense, even if it returns an error. e.g. caller code could try one thing, then if it fails try another (assuming that no state changes were made, even in memory, by the first op)

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -112,12 +112,36 @@ impl FrcXXXToken<RuntimeError> for BasicToken<'_> {
...
+        /*
+        let param_bytes = fvm_ipld_encoding::to_vec(&receiver_params);
+        let message = format!("receiver hook params: {:?}", param_bytes);
+        sdk::vm::abort(fvm_shared::error::ExitCode::USR_ILLEGAL_ARGUMENT.value(),
+                        Some(message.as_str()));
+        */
+
+        if let Err(e) = self.util.call_receiver_hook(&params.initial_owner, receiver_params) {
+            // revert to previous on-chain state
+            sdk::sself::set_root(&old_cid).unwrap();
```
anorth, line 136: We shouldn't need to do this so long as invoke will abort if this function returns an error. I think that's a reasonable API – if any method returns an error, invoke will abort

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -112,12 +112,36 @@ impl FrcXXXToken<RuntimeError> for BasicToken<'_> {
...
+            // need some way to revert the in-memory state
+            // maybe a separate Token function would do it
+            // so we'd have like Token.replace(cloned_data)
```
anorth, line 139: It depends on the lifetime of the Token. In built-in actors, it's not a struct member, but ephemerally constructed for each message. So we just drop it.

anorth (commented):
General direction seems sound, but I think returning an opaque callable rather than requiring the caller to invoke call_receiver_hook would be a better API, plus bonus that we can implement some dynamic checking that it's called.

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
abright: Makes sense, I was thinking about restoring things in order to take other actions (like calling another actor) upon failure. But the more I think about that kind of use, the more it feels like it belongs in a separate actor. Easier at that level to call a method like mint or transfer on the token actor and then take further action based on the result. Also makes this code a lot simpler because it can just return a `Result` and we'll ultimately handle that in `invoke`.

fil_fungible_token/src/token/mod.rs
abright: It didn't seem so useful to me because my original design was going to move the entire mint call into a similar thing. Also wanted to have a "save but revert on failure" version for the receiver hook call, but I think that stuff needs to belong to the state as it's not limited to Token API use (it'd build on my idea from https://github.com/helix-onchain/filecoin/pull/55#discussion_r948616517 ).

I don't think any of this is needed in the abort-on-failure design so I'll remove the comment here and leave the transaction alone.

abright:
This updated version should be a bit cleaner and doesn't bother trying to restore state if minting or the receiver hook call fails - we'll just abort.

fil_fungible_token/src/token/mod.rs
anorth: It's a good point for discussion whether we need this transactional behaviour in the state object, or only in the token object. I could easily be convinced of the latter, but let's defer for another PR.

fil_fungible_token/src/token/mod.rs
```diff
@@ -623,6 +614,32 @@ where
         }
     }
 
+    /// Calls the receiver hook, do not revert on error
+    /// Error results from this should be handled upstream
```
anorth, line 618: I think "do not revert on error" doesn't make much sense to specify here. We can just say it calls the receiver hook, and returns an error if it fails, in response to which most callers should abort.

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -144,13 +145,30 @@ impl Cbor for MintReturn {}
 
 impl BasicToken<'_> {
     fn mint(&mut self, params: MintParams) -> Result<MintReturn, RuntimeError> {
-        self.util.mint(
+        // Token::mint shouldn't save state at all
+        //let old_cid = sdk::sself::root().unwrap();
```
anorth, line 149: Delete these lines. The comment is a bit off given the saving at line 159. But I agree we don't need the old CID, just return an error.

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -144,13 +145,30 @@ impl Cbor for MintReturn {}
...
+        let param_bytes = fvm_ipld_encoding::to_vec(&receiver_params);
+        let message = format!("receiver hook params: {:?}", param_bytes);
+        sdk::vm::abort(fvm_shared::error::ExitCode::USR_ILLEGAL_ARGUMENT.value(),
+                        Some(message.as_str()));
+        */
+
+        self.util.call_receiver_hook(&params.initial_owner, receiver_params)?;
+        // save state and set root here
+        // then call the receiver hook
+        // return ok or err based on that result
```
anorth, line 171: Out of date comment?

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -269,7 +291,7 @@ pub fn invoke(params: u32) -> u32 {
                 }
                 _ => {
                     sdk::vm::abort(
-                        fvm_shared::error::ExitCode::USR_ILLEGAL_ARGUMENT.value(),
+                        ExitCode::USR_ILLEGAL_ARGUMENT.value(),
```
anorth, line 294: USR_UNHANDLED_MESSAGE is for this

anorth (approved):
Looks good, although I would prefer to update the tests at the same time if possible. Would like @alexytsu to also 👍

alexytsu:
> Looks good, although I would prefer to update the tests at the same time if possible. Would like @alexytsu to also 👍

Discussed this with @abright and we thought it would make sense for the new behaviour to be tested at the integration level as it would be awkward to have unit tests assert behaviour which happens across multiple library methods.

Andrew pointed out also that we might get better testing here if we eventually had the `mint`/`transfer` return callables (so that we could assert in unit tests that they tracked whether they were called successfully or not). I'm happy with the updated tests to come in a followup

alexytsu (approved):
LGTM

anorth:
> we thought it would make sense for the new behaviour to be tested at the integration level as it would be awkward to have unit tests assert behaviour which happens across multiple library methods.

Sure, only test what we need to, but let's not have ignored tests hanging around.

Issue #81, opened 2022-08-28
Fix ignored token library tests


anorth:
Some token library tests were ignored in #55 when the token receiver hook call was extracted. Move or rework the tests so that none are ignored.

abright:
They're back in https://github.com/helix-onchain/filecoin/pull/82

Issue #80, opened 2022-08-25
More tests on example token actor


abright:
Currently it only has a single integration test that mints tokens, this should be extended (or new tests added) to involve allowances, transfers and burn so all the major operations are covered. Need to be careful because those tests are quite heavy and could lead to even longer CI runs.

anorth:
Some scenarios to integration-test:
- Mint to A who rejects hook
- Mint to token itself (rejected)

- A transfers to token itself (should be rejected)
- A transfers to A (zero, non-zero)
- A transfers to A and rejects hook
- A transfers to B (zero, non-zero, accept, reject)

- Mint to A, hook burns the tokens
- Mint to A, hook transfers to B
- Mint to A, hook transfers to B who rejects (but A keeps the tokens)

- A transfer to B, hook transfer to C
- A transfer to B, hook transfer to C, hook burns
- A transfer to B, hook transfer to C (rejected), B hook then transfers to D
- A transfer to B, hook transfer back to A (accept / reject)

- A transfer to B, hook burns but then aborts (tokens don't move)
- A transfer to B, hook transfer to C, but then B hook aborts (no move)

Assert all the expected token balances and token supply after each of these.

anorth:
@abright could you update this issue with which, if any, remain as further tests we could write?

abright:
Current state:
- [x] Mint to A who rejects hook
- [x] Mint to token itself (rejected)
- [x] A transfers to token itself (should be rejected)
- [x] A transfers to A (zero, non-zero)
- [ ] A transfers to A and rejects hook
- [x] A transfers to B (zero, non-zero, accept, reject)
- [x] Mint to A, hook burns the tokens
- [x] Mint to A, hook transfers to B
- [x] Mint to A, hook transfers to B who rejects (but A keeps the tokens)
- [x] A transfer to B, hook transfer to C
- [x] A transfer to B, hook transfer to C, hook burns
- [ ] A transfer to B, hook transfer to C (rejected), B hook then transfers to D
- [ ] A transfer to B, hook transfer back to A (accept / reject)
- [ ] A transfer to B, hook burns but then aborts (tokens don't move)
- [ ] A transfer to B, hook transfer to C, but then B hook aborts (no move)

Wasn't too hard to add the missing ones so I've done them in #193

abright:
PR #193 is merged now which covers all the tests outlined here.

PR #82, opened 2022-08-29
Receiver hook call guard for mint/transfer

abright:
This is part 2 of splitting the receiver hook calls out of mint and transfer operations. Implements a callable struct to be returned by mint and transfer functions which holds parameters for the hook call as well as the return data to be handed off after the receiver hook call succeeds. Guards against half-finished transactions by panicking on drop if the hook hasn't been called.

TODO:
- [x] fix tests - see https://github.com/helix-onchain/filecoin/issues/81
- [ ] ~add a transfer integration test - see https://github.com/helix-onchain/filecoin/issues/80~ will do in a new branch
- [x] add doc comments on the `ReceiverHook` type

### DIFF
--- a/fil_fungible_token/src/receiver/mod.rs
+++ b/fil_fungible_token/src/receiver/mod.rs
@@ -71,3 +71,42 @@ impl std::ops::Drop for ReceiverHook {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use fvm_ipld_encoding::RawBytes;
+    use fvm_shared::{address::Address, econ::TokenAmount};
+    use num_traits::Zero;
+
+    use super::{types::TokensReceivedParams, ReceiverHook};
+    use crate::runtime::messaging::FakeMessenger;
+
+    const TOKEN_ACTOR: Address = Address::new_id(1);
+    const ALICE: Address = Address::new_id(2);
+
+    fn generate_hook() -> ReceiverHook {
+        let params = TokensReceivedParams {
+            operator: TOKEN_ACTOR.id().unwrap(),
+            from: TOKEN_ACTOR.id().unwrap(),
+            to: ALICE.id().unwrap(),
+            amount: TokenAmount::zero(),
+            operator_data: RawBytes::default(),
+            token_data: RawBytes::default(),
+        };
+        ReceiverHook::new(ALICE, params)
+    }
+
+    #[test]
+    fn calls_hook() {
+        let mut hook = generate_hook();
+        let msg = FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 3);
+        hook.call(&msg).unwrap();
+    }
+
+    #[test]
+    #[should_panic]
+    fn panics_if_not_called() {
+        let mut _hook = generate_hook();
+        // _hook should panic when dropped as we haven't called the hook
+    }
+}
--- a/fil_fungible_token/src/receiver/mod.rs
+++ b/fil_fungible_token/src/receiver/mod.rs
@@ -1,1 +1,56 @@
+use fvm_ipld_encoding::RawBytes;
+use fvm_shared::{address::Address, econ::TokenAmount, error::ExitCode};
+use num_traits::Zero;
+use types::TokensReceivedParams;
+
+use crate::runtime::messaging::{Messaging, RECEIVER_HOOK_METHOD_NUM};
+use crate::token::TokenError;
+
 pub mod types;
+
+#[derive(Debug)]
+pub struct ReceiverHookGuard<T> {
+    address: Address,
+    params: Option<TokensReceivedParams>,
+    return_value: Option<T>,
+}
+
+impl<T> ReceiverHookGuard<T> {
+    pub fn new(address: Address, params: TokensReceivedParams, return_value: T) -> Self {
+        ReceiverHookGuard { address, params: Some(params), return_value: Some(return_value) }
+    }
+    pub fn call(&mut self, msg: &dyn Messaging) -> std::result::Result<T, TokenError> {
+        if self.params.is_none() {
+            return Err(TokenError::ReceiverHookGuardAlreadyCalled);
+        }
+
+        // this will leave self.params set to None, so a further attempt to call() will fail
+        let params = self.params.take().unwrap();
+
+        let receipt = msg.send(
+            &self.address,
+            RECEIVER_HOOK_METHOD_NUM,
+            &RawBytes::serialize(&params)?,
+            &TokenAmount::zero(),
+        )?;
+
+        match receipt.exit_code {
+            ExitCode::OK => Ok(self.return_value.take().unwrap()),
+            abort_code => Err(TokenError::ReceiverHook {
+                from: params.from,
+                to: params.to,
+                operator: params.operator,
+                amount: params.amount,
+                exit_code: abort_code,
+            }),
+        }
+    }
+}
+
+impl<T> std::ops::Drop for ReceiverHookGuard<T> {
+    fn drop(&mut self) {
+        if self.params.is_some() {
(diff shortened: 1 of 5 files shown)


fil_fungible_token/src/token/mod.rs
```diff
@@ -118,7 +123,7 @@ where
         amount: &TokenAmount,
         operator_data: RawBytes,
         token_data: RawBytes,
-    ) -> Result<TokensReceivedParams> {
+    ) -> Result<ReceiverHookGuard<TokenAmount>> {
```
abright, line 126: I think this `TokenAmount` (total supply) should be either wrapped in a struct again, or documented in the comments.

fil_fungible_token/src/receiver/mod.rs
```diff
@@ -1 +1,56 @@
...
+    pub fn new(address: Address, params: TokensReceivedParams, return_value: T) -> Self {
+        ReceiverHookGuard { address, params: Some(params), return_value: Some(return_value) }
+    }
+    pub fn call(&mut self, msg: &dyn Messaging) -> std::result::Result<T, TokenError> {
+        if self.params.is_none() {
+            return Err(TokenError::ReceiverHookGuardAlreadyCalled);
+        }
+
+        // this will leave self.params set to None, so a further attempt to call() will fail
+        let params = self.params.take().unwrap();
```
anorth, line 28: I find this use of an option to be unnecessarily obscure, vs an explicit "was_called" field. I don't think we're saving any memory or anything vs that approach.

fil_fungible_token/src/receiver/mod.rs
```diff
@@ -1 +1,56 @@
...
+                exit_code: abort_code,
+            }),
+        }
+    }
+}
+
+impl<T> std::ops::Drop for ReceiverHookGuard<T> {
+    fn drop(&mut self) {
+        if self.params.is_some() {
+            panic!("dropped before receiver hook was called");
```
anorth, line 53: Please include all the information we can in this panic message, e.g. the address that was to be called

fil_fungible_token/src/receiver/mod.rs
```diff
@@ -1 +1,56 @@
...
+use crate::runtime::messaging::{Messaging, RECEIVER_HOOK_METHOD_NUM};
+use crate::token::TokenError;
+
 pub mod types;
+
+#[derive(Debug)]
+pub struct ReceiverHookGuard<T> {
+    address: Address,
+    params: Option<TokensReceivedParams>,
+    return_value: Option<T>,
```
anorth, line 15: I don't think threading this otherwise-unrelated value here is good. I see how in the case that calling the hook is exactly the last thing the caller does that it can make for a concise syntax, but that's also potentially confusing about what's being returned.

Focus this type just on the hook thunk. The mint/transfer methods can return an anonymous tuple of the result and this hook thunk.

fil_fungible_token/src/receiver/mod.rs
```diff
@@ -1 +1,56 @@
...
+use num_traits::Zero;
+use types::TokensReceivedParams;
+
+use crate::runtime::messaging::{Messaging, RECEIVER_HOOK_METHOD_NUM};
+use crate::token::TokenError;
+
 pub mod types;
+
+#[derive(Debug)]
+pub struct ReceiverHookGuard<T> {
```
anorth, line 12: I would just call this `ReceiverHook`, or maybe `ReceiverHookThunk`, tho that might be too much jargon. Please add documentation comments explaining its behaviour.

fil_fungible_token/src/token/error.rs
```diff
@@ -31,6 +31,8 @@ pub enum TokenError {
         amount: TokenAmount,
         exit_code: ExitCode,
     },
+    #[error("receiver hook was called already")]
+    ReceiverHookGuardAlreadyCalled,
```
anorth, line 35:
```suggestion
    ReceiverHookAlreadyCalled,
```

fil_fungible_token/src/token/mod.rs
anorth: I have landed a `MintReturn` on master already

fil_fungible_token/src/receiver/mod.rs
abright: I'll remove it then. It will look a little more awkward in use:
```Rust
let (mut hook, ret) = token.mint(...)?;
```
But it does give you access to the return value immediately.

The original concept was that the operation and receiver hook call should be seen as two halves of the one transaction, so the final result was returned by the hook call.

fil_fungible_token/src/receiver/mod.rs
abright: Previous iterations required it because of ownership and lifetime concerns. I don't think there's anything special in the `TokensReceivedParams` that will stand in the way now, so I'll switch it over.

fil_fungible_token/src/receiver/mod.rs
abright: Including debug formatting of the target address and params is easy enough:
```
"panicked at 'dropped before receiver hook was called on Address { network: Mainnet, payload: ID(10010) } with TokensReceivedParams { from: 10000, to: 10010, operator: 100, amount: 100, operator_data: RawBytes { bytes: [] }, token_data: RawBytes { bytes: [] } }', /home/abright/projects/filecoin/fil_fungible_token/src/receiver/mod.rs:52:13"
```
This might be a bit noisy but it should (hopefully) only be seen during development

abright:
I'm working on another integration test to do some transfers as well as minting, but it might be better to work on that in a separate branch as this already changes a lot of code, especially in the tests.

fil_fungible_token/src/token/mod.rs
```diff
@@ -1043,8 +1071,9 @@ mod test {
         };
 
         // state remained unchanged
-        assert_eq!(token.balance_of(TREASURY).unwrap(), TokenAmount::zero());
-        assert_eq!(token.total_supply(), TokenAmount::zero());
+        // TODO: we abort rather than manually revert state, probably need to get rid of these checks
+        //assert_eq!(token.balance_of(TREASURY).unwrap(), TokenAmount::zero());
+        //assert_eq!(token.total_supply(), TokenAmount::zero());
```
abright, line 1076: Should this just be removed now that the state isn't reverted when receiver hook aborts? I think checking this post-failure state now needs to be done in an integration test

alexytsu: Will the library give any utilities to assist the state revert? If so, it can be tested here ("mocking" the expected call). 

Else, I think it can only be tested at an actor level

abright: Added a `replace` function to `Token` to do it, so we can clone the original state and swap over to that in place of the modified state where tests need us to revert. Trying to use it outside of tests would be a swift trip to borrow-checker hell so it's limited by config and visibility flags to tests only.

fil_fungible_token/src/receiver/mod.rs
```diff
@@ -1 +1,73 @@
...
+/// Drop implements the panic if not called behaviour
+impl std::ops::Drop for ReceiverHook {
+    fn drop(&mut self) {
+        if !self.called {
+            panic!(
+                "dropped before receiver hook was called on {:?} with {:?}",
+                self.address, self.params
+            );
+        }
+    }
```
anorth, line 72: Please add a unit test for this behaviour. And possibly the normal path too, though it's adequately covered by token lib tests.

anorth (approved):

abright:
Not sure why CI decided to pick a fight today, it's breaking on the backtrace stuff in `fvm` which should no longer be an issue

PR #84, opened 2022-08-29
Tweaks to return types etc.

anorth:
Adds a return type for `mint`, and type aliases for the bare token amounts that are returned.


alexytsu (approved):
LGTM - thanks

Issue #85, opened 2022-08-29
Add granularity method


anorth:
See the FRC. Add the method to the library, and appropriate types/aliases, and to the example actor.

PR #86, opened 2022-08-30
add granularity

alexytsu:
https://github.com/helix-onchain/filecoin/issues/85

Adds granularity to library. Since granularity cannot change, it has not been added to state, instead it should be defined statically similar to name and symbol

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -2195,7 +2195,6 @@ mod test {
         );
 
         assert_eq!(token.granularity(), 100);
-        assert_eq!(token.granularity(), 100);
 
         // Minting
         token
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -122,6 +122,14 @@ where
     BS: Blockstore,
     MSG: Messaging,
 {
+    /// Returns the smallest amount of tokens which is indivisible
+    ///
+    /// Transfers and balances must be in multiples of granularity but allowances need not be.
+    /// Granularity never changes after it is initially set
+    pub fn granularity(&self) -> u64 {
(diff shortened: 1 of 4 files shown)


fil_fungible_token/src/token/types.rs
```diff
@@ -98,6 +107,7 @@ pub trait FRC46Token<E> {
     fn burn_from(&mut self, params: BurnFromParams) -> Result<BurnFromReturn, E>;
 }
 
+type GranularityReturn = u64;
 type TotalSupplyReturn = TokenAmount;
 type BalanceReturn = TokenAmount;
 type AllowanceReturn = TokenAmount;
```
alexytsu, line 113: Should these be made `pub` so that they can be used in actor code?

fil_fungible_token/src/token/mod.rs
```diff
@@ -2074,8 +2082,17 @@ mod test {
...
+        // construct token with 100 granularity
+        let mut token = Token::wrap(
+            bs,
+            FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 6),
+            100,
+            &mut token_state,
+        );
+
+        assert_eq!(token.granularity(), 100);
+        assert_eq!(token.granularity(), 100);
```
abright, line 2095: Duplicated checks?

fil_fungible_token/src/token/types.rs
abright: They are aliases so you can mix and match them with the underlying type (compiler doesn't mind either way). I think it's worth exposing them so we can use in our example code, Though the underlying types also seem obvious enough for things like granularity or total supply.

anorth: Yes please do.

anorth (approved):

PR #93, opened 2022-09-02
Universal token receiver

abright:
Implements the universal receiver hook as detailed in [FRC-0046](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0046.md)

Also makes a change to our `FvmMessenger` which fixes an issue I had with actor-to-actor messaging. The receiver actor example can now deserialise and inspect details of incoming transfers

### DIFF
--- a/fil_fungible_token/src/receiver/types.rs
+++ b/fil_fungible_token/src/receiver/types.rs
@@ -1,3 +1,4 @@
+use frc42_dispatch::method_hash;
 use fvm_ipld_encoding::tuple::{Deserialize_tuple, Serialize_tuple};
 use fvm_ipld_encoding::{Cbor, RawBytes};
 use fvm_shared::bigint::bigint_ser;
@@ -12,11 +13,14 @@ pub trait FRC46TokenReceiver {
     /// the receiving actor can immediately utilise the received funds. If the receiver wishes to
     /// reject the incoming transfer, this function should abort which will cause the token actor
     /// to rollback the transaction.
-    fn tokens_received(params: TokensReceivedParams);
+    fn receive(token_type: TokenType, params: FRC46TokenReceived);
 }
 
+pub type TokenType = u32;
+pub const FRC46_TOKEN_TYPE: TokenType = method_hash!("FRC46") as u32;
+
 #[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
-pub struct TokensReceivedParams {
+pub struct FRC46TokenReceived {
     /// The account that the tokens are being pulled from (the token actor address itself for mint)
     pub from: ActorID,
     /// The account that the tokens are being sent to (the receiver address)
@@ -31,4 +35,11 @@ pub struct TokensReceivedParams {
     pub token_data: RawBytes,
 }
 
-impl Cbor for TokensReceivedParams {}
+impl Cbor for FRC46TokenReceived {}
+
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct ReceiveParams {
+    pub type_: TokenType,
+    pub payload: RawBytes,
+}
+impl Cbor for ReceiveParams {}
--- a/fil_fungible_token/src/receiver/types.rs
+++ b/fil_fungible_token/src/receiver/types.rs
@@ -5,20 +5,22 @@ use fvm_shared::bigint::bigint_ser;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::ActorID;
 
-/// Standard interface for an actor that wishes to receive FRC-XXX tokens
-pub trait FRC46TokenReceiver {
+/// Standard interface for an actor that wishes to receive FRC-XXX tokens or other assets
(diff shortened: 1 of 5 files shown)


fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -71,8 +71,7 @@ impl Messaging for FvmMessenger {
         params: &RawBytes,
         value: &TokenAmount,
     ) -> Result<Receipt> {
-        let params = RawBytes::new(fvm_ipld_encoding::to_vec(&params)?);
-        Ok(send::send(to, method, params, value.clone())?)
+        Ok(send::send(to, method, params.clone(), value.clone())?)
```
alexytsu, line 74: 👍 thanks

alexytsu (approved):
LGTM

fil_fungible_token/src/runtime/messaging.rs
abright: I want to have a further look at this later.  I think every time we call this, we're creating a `RawBytes` at the call site to serialise some params struct. Might be better to just move ownership (of the `RawBytes`) in that case to avoid the clone and save a little gas.

fil_fungible_token/src/receiver/types.rs
```diff
@@ -12,11 +13,14 @@ pub trait FRC46TokenReceiver {
     /// the receiving actor can immediately utilise the received funds. If the receiver wishes to
     /// reject the incoming transfer, this function should abort which will cause the token actor
     /// to rollback the transaction.
-    fn tokens_received(params: TokensReceivedParams);
+    fn receive(token_type: TokenType, params: FRC46TokenReceived);
```
anorth, line 16: Perhaps generalise this to be a UniversalReceiver trait, and with the params argument here being raw bytes. We're going to want essentially the same thing for NFT (so perhaps this will move out of this crate to a new fil_universal_receiver crate too). 

But you could convince me that being generic for all receivers is out of scope here.

fil_fungible_token/src/receiver/types.rs
```diff
@@ -12,11 +13,14 @@ pub trait FRC46TokenReceiver {
...
 }
 
+pub type TokenType = u32;
```
anorth, line 19: Change to `ReceiverType` or similar. The universal receiver isn't limited to "tokens".

fil_fungible_token/src/receiver/types.rs
```diff
@@ -31,4 +35,11 @@ pub struct TokensReceivedParams {
     pub token_data: RawBytes,
 }
 
-impl Cbor for TokensReceivedParams {}
+impl Cbor for FRC46TokenReceived {}
+
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct ReceiveParams {
```
anorth, line 41: 👍 . UniversalReceiverParams?

fil_fungible_token/src/token/mod.rs
```diff
@@ -694,9 +694,10 @@ mod test {
         Token::wrap(bs, FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 6), 1, state)
     }
 
-    fn assert_last_hook_call_eq(messenger: &FakeMessenger, expected: TokensReceivedParams) {
+    fn assert_last_hook_call_eq(messenger: &FakeMessenger, expected: FRC46TokenReceived) {
         let last_called = messenger.last_message.borrow().clone().unwrap();
-        let last_called: TokensReceivedParams = last_called.deserialize().unwrap();
+        let last_called: ReceiveParams = last_called.deserialize().unwrap();
+        let last_called: FRC46TokenReceived = last_called.payload.deserialize().unwrap();
```
anorth, line 700: Check the token type too

fil_fungible_token/src/runtime/messaging.rs
```diff
@@ -52,9 +52,9 @@ pub trait Messaging {
     fn initialize_account(&self, address: &Address) -> Result<ActorID>;
 }
 
-// the method number comes from taking the name as "TokensReceived" and applying
+// the method number comes from taking the name as "Received" and applying
 // the transformation described in https://github.com/filecoin-project/FIPs/pull/399
-pub const RECEIVER_HOOK_METHOD_NUM: u64 = method_hash!("TokensReceived");
+pub const RECEIVER_HOOK_METHOD_NUM: u64 = method_hash!("Received");
```
anorth, line 57:
```suggestion
pub const RECEIVER_HOOK_METHOD_NUM: u64 = method_hash!("Receive");
```

anorth (approved):

fil_fungible_token/src/receiver/types.rs
anorth: Moving things around can wait for a follow-up of course

abright: Updated to this: 
```Rust
fn receive(params: UniversalReceiverParams);
```

I've also added more comments on it and the `UniversalReceiverParams` which hopefully explain things clearly

PR #96, opened 2022-09-07
NFT API

alexytsu:
Raising a PR for some early work in progress code for NFTs. The implementation details here are not final and will change but the overall structure of the code and testing structure is worth merging. 

I'll be developing this library with integration tests from the get go to find issues with library usability earlier rather than later: so hopefully a bit smoother integration than fungible_tokens.

Also, I'm expecting some methods/functions might have multiple implementations in early stages (e.g. batched vs unbatched impls) so we can compare gas costs and patterns via the integration tests.


frcxx_nft/src/nft/state.rs
```diff
@@ -0,0 +1,98 @@
...
+    IpldAmt(#[from] AmtError),
+    #[error("other error: {0}")]
+    Other(String),
+}
+
+/// NFT state IPLD structure
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct NFTSetState {
+    /// Amt<ActorID> of balances as a Hamt where index is TokenID
+    pub tokens: Cid,
```
alexytsu, line 32: toy implementation:
- will be better represented as a struct or bitfield that encodes other information such as ownership, operators etc.

testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
```diff
@@ -0,0 +1,76 @@
+use fil_fungible_token::runtime::blockstore::Blockstore;
```
alexytsu, line 1: Currently fil_fungible_token implements a SDK-delegated blockstore. 

- We could reimplement (copy-paste) this blockstore in the frcxx_nft library. But that brings the dependency on `anyhow` into our library code
- Or perhaps doesn't need to be in either of our libraries at all? It's a minimal interface to implement so can be done purely in actor example code
- We could even just import a library such https://github.com/polyphene/fvm-rs-sdk/blob/main/sdk/src/state/cbor.rs that implements it

frcxx_nft/src/nft/state.rs
```diff
@@ -0,0 +1,98 @@
...
+use fvm_ipld_encoding::CborStore;
+use fvm_ipld_encoding::DAG_CBOR;
+use fvm_shared::ActorID;
+use thiserror::Error;
+
+pub type TokenID = u64;
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
+pub struct BatchMintReturn {
+    pub tokens: Vec<TokenID>,
```
anorth, line 17: The Bitfield structure will really come into its own here, I recommend we use it (and spec it to be used in the FRC)

frcxx_nft/src/nft/state.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[error("ipld hamt error: {0}")]
+    IpldAmt(#[from] AmtError),
+    #[error("other error: {0}")]
+    Other(String),
+}
+
+/// NFT state IPLD structure
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct NFTSetState {
+    /// Amt<ActorID> of balances as a Hamt where index is TokenID
```
anorth, line 31: Clarify `Amt<TokenID, ActorId>`

frcxx_nft/src/nft/state.rs
anorth, line 32: Maybe `owners`? `owners_by_token`? `token_owners`? Others might also have an owner->tokens map, which one might call `tokens`. Tho both ways can be confused.

frcxx_nft/src/nft/state.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    fn get_token_amt<'bs, BS: Blockstore>(&self, store: &'bs BS) -> Result<Amt<ActorID, &'bs BS>> {
+        let res = Amt::load(&self.tokens, store)?;
+        Ok(res)
+    }
+
+    pub fn mint_token<BS: Blockstore>(&mut self, bs: &BS, owner: ActorID) -> Result<TokenID> {
+        let mut token_map = self.get_token_amt(&bs)?;
+        let new_index = token_map.count();
```
anorth, line 76: This is invalid after burning (assuming you delete the item rather than tombstone). I think you should store next_id explicitly. Also need a total_supply, since AMT doesn't implement a `size()` field

frcxx_nft/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+
+pub trait FRCXXXNFT {
+    /// A descriptive name for the collection of NFTs in this actor
+    fn name(&self) -> String;
+
+    /// An abbreviated name for NFTs in this contract
+    fn symbol(&self) -> String;
+
+    /// Gets a link to associated metadata for a given NFT
+    fn metadata_uri(&self, params: TokenID) -> Cid;
```
anorth, line 17: If we're using CIDs, it won't be a URI. MetadataID perhaps?

frcxx_nft/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Gets the total number of NFTs in this actor
+    fn total_supply(&self) -> u64;
+
+    /// Burns a given NFT, removing it from the total supply and preventing new NFTs from being
+    /// minted with the same ID
+    fn burn(&self, params: TokenID);
+
+    /// Gets a list of all the tokens in the collection
+    /// FIXME: make this paginated
+    fn list_tokens(&self) -> Vec<TokenID>;
```
anorth, line 28: And use bitfield

frcxx_nft/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// Query if the address is the approved operator for another address
+    fn is_approved_for_all(&self, params: IsApprovedForAllParams) -> bool;
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
+pub struct TransferParams {
+    pub from: Address,
+    pub to: Address,
+    pub token_ids: Vec<TokenID>,
```
anorth, line 60: Bitfields everywhere here

frcxx_nft/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Get the approved address for a single NFT
+    fn get_approved(&self, params: TokenID) -> ActorID;
+
+    /// Query if the address is the approved operator for another address
+    fn is_approved_for_all(&self, params: IsApprovedForAllParams) -> bool;
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
+pub struct TransferParams {
+    pub from: Address,
```
anorth, line 58: `from` should be absent here, implicitly the caller.

frcxx_nft/src/types.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub operator_data: RawBytes,
+}
+
+impl Cbor for TransferParams {}
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
+pub struct TransferFromParams {
+    pub from: Address,
+    pub to: Address,
+    pub operator: Address,
```
anorth, line 70: Delete operator, the explicit `from` is what makes the difference here

testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
anorth: Wanna see if you can go fix FVM to remove anyhow from this? It might not be easy tho

testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+use frcxx_nft::nft::state::{BatchMintReturn, NFTSetState};
+use fvm_ipld_encoding::{de::DeserializeOwned, ser, RawBytes, DAG_CBOR};
+use fvm_sdk as sdk;
+use fvm_shared::error::ExitCode;
+use sdk::NO_DATA_BLOCK_ID;
+
+/// Grab the incoming parameters and convert from RawBytes to deserialized struct
+pub fn deserialize_params<O: DeserializeOwned>(params: u32) -> O {
+    let params = sdk::message::params_raw(params).unwrap().1;
```
anorth, line 10: We should really return a serialisation error rather than panic here

testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+    let params = RawBytes::new(params);
+    params.deserialize().unwrap()
+}
+
+fn return_ipld<T>(value: &T) -> u32
+where
+    T: ser::Serialize + ?Sized,
+{
+    let bytes = fvm_ipld_encoding::to_vec(value).unwrap();
```
anorth, line 19: Ditto

anorth (commented):
Looks like a good start to me. Do you want approval to land this as WIP?

alexytsu:
> Looks like a good start to me. Do you want approval to land this as WIP?

@anorth Yes - pls. I'll make the suggested changes and add new features in follow ups. 

I'd like this boilerplate out of the way of those diffs and the ability to test any changes end-to-end

anorth (approved):

PR #97, opened 2022-09-07
upgrade token amount

alexytsu:
updates fvm_shared to use the new `TokenAmount`

this forces upgrade of fvm_sdk, fvm and fvm_integration_tests

should also unblock using blake2b512 for the hasher


anorth (approved):
Thanks!

PR #98, opened 2022-09-08
Expose more functionality from the token library

alexytsu:
Makes the library a bit more flexible to be useful outside of straightforward FRC-46 token actors

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -294,6 +294,35 @@ where
         Ok(())
     }
 
+    /// Sets the allowance to a specified amount
+    pub fn set_allowance(
+        &mut self,
+        owner: &Address,
+        operator: &Address,
+        amount: &TokenAmount,
+    ) -> Result<()> {
+        let owner = match self.get_id(owner) {
+            Ok(owner) => owner,
+            Err(MessagingError::AddressNotResolved(_)) => {
+                // uninitialized address has implicit zero allowance already
+                return Ok(());
+            }
+            Err(e) => return Err(e.into()),
+        };
+        let operator = match self.get_id(operator) {
+            Ok(operator) => operator,
+            Err(MessagingError::AddressNotResolved(_)) => {
+                // uninitialized address has implicit zero allowance already
+                return Ok(());
+            }
+            Err(e) => return Err(e.into()),
+        };
+        // if both accounts resolved, explicitly set allowance to zero
+        self.state.set_allowance(&self.bs, owner, operator, amount)?;
+
+        Ok(())
+    }
+
     /// Burns an amount of token from the specified address, decreasing total token supply
     ///
     /// - The requested value MUST be non-negative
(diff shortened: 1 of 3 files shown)


fil_fungible_token/src/token/state.rs
```diff
@@ -317,6 +360,42 @@ impl TokenState {
...
+        bs: &BS,
+        owner: ActorID,
+        operator: ActorID,
+        amount: &TokenAmount,
+    ) -> Result<TokenAmount> {
+        if amount.is_negative() {
+            return Err(StateError::NegativeAllowance { owner, operator, amount: amount.clone() });
+        }
+
+        let old_allowance = self.get_allowance_between(bs, owner, operator)?;
```
anorth, line 375: Get this value fro the allowance_map that you're loading below anyway.

fil_fungible_token/src/token/mod.rs
```diff
@@ -294,6 +304,32 @@ where
...
 
+    /// Sets the allowance to a specified amount, returning the old allowance
+    pub fn set_allowance(
+        &mut self,
+        owner: &Address,
+        operator: &Address,
+        amount: &TokenAmount,
+    ) -> Result<TokenAmount> {
+        if amount.is_negative() {
+            return Err(TokenError::InvalidNegative { name: "allowance", amount: amount.clone() });
```
anorth, line 315: The state method already does this.

fil_fungible_token/src/token/mod.rs
```diff
@@ -294,6 +304,32 @@ where
...
+        }
+
+        // Handle special revoke allowance case
+        if amount.is_zero() {
+            let old_allowance = self.allowance(owner, operator)?;
+            self.revoke_allowance(owner, operator)?;
+            return Ok(old_allowance);
+        }
```
anorth, line 323: 1) I think you need to resolve the addresses before looking up the allowance (can you write a test that fails here?)
2) The state method already does this, why is it repeated here?

fil_fungible_token/src/token/mod.rs
```diff
@@ -237,7 +237,12 @@ where
         operator: &Address,
         delta: &TokenAmount,
     ) -> Result<TokenAmount> {
-        let delta = validate_amount(delta, "allowance delta", self.granularity)?;
+        if delta.is_negative() {
+            return Err(TokenError::InvalidNegative {
+                name: "allowance delta",
+                amount: delta.clone(),
+            });
+        }
```
anorth, line 245: Just to confirm, did you remove the granularity check here because the FRC says so? Ok.

I think we should pull this out to a `validate_allowance` method so it's clear that this isn't an oversight. I can see some well-meaning future person trying to change this back.

fil_fungible_token/src/token/state.rs
```diff
@@ -187,14 +191,53 @@ impl TokenState {
...
         &self,
         bs: &'bs BS,
     ) -> Result<Map<'bs, BS, ActorID, TokenAmount>> {
         Ok(Hamt::load_with_bit_width(&self.balances, bs, HAMT_BIT_WIDTH)?)
     }
 
+    /// Retrieve the number of token holders
+    ///
+    /// This involves iterating through the entire HAMT
+    pub fn get_number_token_holders<BS: Blockstore>(&self, bs: &BS) -> Result<usize> {
```
anorth, line 236: `get_balance_count`? `balance_count`? `count_balances` (sounds slow, like it is)?

fil_fungible_token/src/token/mod.rs
alexytsu: I've removed it from the library level and let the state level handle this

fil_fungible_token/src/token/mod.rs
alexytsu: `revoke_allowance` and `allowance` have special shortcut cases which avoid initialising accounts that don't need to be initialised. So I delegate the logic of checking that to the existing methods. 

however, this means in cases where we know amount is being set to non-zero we have a duplicate sign check at the library and state level. I think this is ok as a defensive measure - since other sites can call set_allowance without checking sign.

alexytsu: an alternative factoring could repeat the special handling of account-initialisation here but it would likely be a bit more verbose.

alexytsu: or - perhaps for functionality beyond the standard interface, we should force consumers to operate directly on a `&mut` of State. this potentially serves also to highlight that extra care should be taken to maintain desired invariants.

fil_fungible_token/src/token/mod.rs
```diff
@@ -294,6 +296,28 @@ where
...
+    pub fn set_allowance(
+        &mut self,
+        owner: &Address,
+        operator: &Address,
+        amount: &TokenAmount,
+    ) -> Result<TokenAmount> {
+        // Handle special revoke allowance case
+        if amount.is_zero() {
+            let old_allowance = self.allowance(owner, operator)?;
+            self.revoke_allowance(owner, operator)?;
```
anorth, line 309: Ok, please document the intent of handling this special case, which is to avoid initialising addresses unnecessarily.

I was confused earlier here and thought this was calling through to a state method, rather than self method. But since we're optimising, change `revoke_allowance` to return the old allowance (via a similar change to the state method). It's already being read and discarded at `Map::delete`.

fil_fungible_token/src/token/mod.rs
```diff
@@ -294,6 +296,28 @@ where
...
+            return Ok(old_allowance);
+        }
+
+        // Attempt to instantiate the accounts if they don't exist
+        let owner = self.resolve_or_init(owner)?;
+        let operator = self.resolve_or_init(operator)?;
+
+        // if both accounts resolved, explicitly set allowance to zero
```
anorth, line 317: "...to zero" is out of place here.

fil_fungible_token/src/token/mod.rs
```diff
@@ -294,6 +296,28 @@ where
         Ok(())
     }
 
+    /// Sets the allowance to a specified amount, returning the old allowance
+    pub fn set_allowance(
+        &mut self,
+        owner: &Address,
+        operator: &Address,
+        amount: &TokenAmount,
+    ) -> Result<TokenAmount> {
```
anorth, line 305: This should `validate_allowance` (because it makes sense, and also consistency with `set_balance`).

anorth (approved):
LGTM if you agree with my remaining suggestions.

Issue #89, opened 2022-08-30
Update FRC42 dispatch to use Blake2b-512 and larger rejection sampling


anorth:
See https://github.com/filecoin-project/FIPs/pull/445

You'll need a new release of the FVM SDK to get the blake-512.

PR #100, opened 2022-09-09
Dispatch update

abright:
See issue https://github.com/helix-onchain/filecoin/issues/89

Updates the dispatch crate family to match latest updates in [FRC-0042](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0042.md)

- [x] Updates the blake2b hash syscall to use the 512-bit version
- [x] Expands the reserved method number range from 8 bits to 24


alexytsu (approved):

anorth (approved):
Thanks.

I wish we had some "golden" tests that needed updating here when the actual values produced by this function change. Such expected values could form part of the FRC spec.

PR #102, opened 2022-09-12
Receiver Hook return data

abright:
Passes through any data returned by the receiver hook

### DIFF
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -163,7 +163,11 @@ where
         let result = self.transaction(|state, bs| {
             let balance = state.change_balance_by(&bs, owner_id, amount)?;
             let supply = state.change_supply_by(amount)?;
-            Ok(MintReturn { balance, supply: supply.clone() })
+            Ok(MintReturn {
+                balance,
+                supply: supply.clone(),
+                hook_return_data: RawBytes::default(),
+            })
         })?;
 
         // return the params we'll send to the receiver hook
--- a/fil_fungible_token/src/token/mod.rs
+++ b/fil_fungible_token/src/token/mod.rs
@@ -785,7 +785,7 @@ mod test {
             &mut actor_state.token_state,
         );
 
-        let (mut hook, _) = token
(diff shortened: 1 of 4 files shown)


fil_fungible_token/src/token/mod.rs
```diff
@@ -155,7 +155,11 @@ where
         let result = self.transaction(|state, bs| {
             let balance = state.change_balance_by(&bs, owner_id, amount)?;
             let supply = state.change_supply_by(amount)?;
-            Ok(MintReturn { balance, supply: supply.clone() })
+            Ok(MintReturn {
+                balance,
+                supply: supply.clone(),
+                hook_return_data: RawBytes::default(),
```
anorth, line 161: In case there are other hooks (sender?) or similar later let's call this `recipient_data`. This parallels `operator_data` and `token_data`, and I think the fact that it's in response/return is clear from context.

fil_fungible_token/src/token/mod.rs
```diff
@@ -425,11 +429,19 @@ where
                     }
                     .into());
                 }
-                Ok(TransferReturn { from_balance: balance.clone(), to_balance: balance })
+                Ok(TransferReturn {
+                    from_balance: balance.clone(),
+                    to_balance: balance,
+                    hook_return_data: RawBytes::default(),
```
anorth, line 435: Ah, yuk, we don't have the value yet. Token devs are going to screw this up.

Ok, in light of this all of a sudden I like your original design of putting the TransferReturn _inside_ the hook object, and returning it only after the hook is invoked.

anorth (changes requested):

fil_fungible_token/src/token/mod.rs
abright: ok, renamed it

fil_fungible_token/src/token/mod.rs
abright: I've changed it to something close to my original design - it's a bit more tricky behind the scenes but should be much nicer for users of the library.

anorth (approved):

PR #95, opened 2022-09-06
Transfer integration test

abright:
Adds a new actor and integration test that mints and transfers tokens

PR #103, opened 2022-09-12
More integration tests

abright:
More tests that build on #95 

Adds a configurable test actor for testing different scenarios such as those outlined in #80 

Single actor tests implemented:
- [x] Mint to A who rejects hook
- [x] Mint to token itself (rejected)
- [x] Mint to A, hook burns the tokens
- [x] A transfers to token itself (should be rejected)
- [x] A transfers to A (zero, non-zero)
- [x] A transfers to A and rejects hook

Multi-actor tests implemented:
- [x] A transfers to B (zero, non-zero, accept, reject)
- [x] Mint to A, hook transfers to B
- [x] Mint to A, hook transfers to B who rejects (but A keeps the tokens)
- [x] A transfer to B, hook transfer to C
- [x] A transfer to B, hook transfer to C, hook burns

Cases from #80 still not implemented:
 - A transfer to B, hook transfer to C (rejected), B hook then transfers to D
 - A transfer to B, hook transfer back to A (accept / reject)
 - A transfer to B, hook burns but then aborts (tokens don't move)
 - A transfer to B, hook transfer to C, but then B hook aborts (no move)

Tests are combined into two files (single- and multi-actor) as a compromise between readability and CI build/run time.

There's also a mechanism to reload the current state from a `Cid` which is used in `Mint`, `Transfer` and `TransferFrom` to update state and then the return value if the state has changed since calling the receiver hook. This covers cases such as transferring or burning from within the receiver hook. It could also be used to revert state if the hook call aborts, which may be useful to some implementations?


testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -198,6 +243,7 @@ pub fn invoke(params: u32) -> u32 {
             let root_cid = sdk::sself::root().unwrap();
 
             let bs = Blockstore::default();
+            // might want these to be done inside each handler as there's times we'll want to discard and reload it
```
abright, line 246: Think I've managed to avoid this, will remove the comment

testing/fil_token_integration/actors/test_actor/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+                    // ignore failures at this level and return the transfer call receipt so caller can decide what to do
+                    return_ipld(&receipt)
+                }
+                TestAction::Burn => {
+                    // burn the tokens
+                    let burn_params = BurnParams {
+                        amount: token_params.amount,
+                    };
+                    let receipt = sdk::send::send(&Address::new_id(sdk::message::caller()), method_hash!("Burn"), RawBytes::serialize(&burn_params).unwrap(), TokenAmount::zero()).unwrap();
+                    if !receipt.exit_code.is_success() {
```
abright, line 121: Not sure if this should explode or just return the result upstream (so failure to burn will keep the tokens) like Transfer does above

testing/fil_token_integration/actors/test_actor/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+                    return_ipld(&receipt)
+                }
+                TestAction::Burn => {
+                    // burn the tokens
+                    let balance = get_balance();
+                    let burn_params = BurnParams {
+                        amount: balance,
+                    };
+                    let receipt = sdk::send::send(&params.token_address, method_hash!("Burn"), RawBytes::serialize(&burn_params).unwrap(), TokenAmount::zero()).unwrap();
+                    if !receipt.exit_code.is_success() {
```
abright, line 170: Same as the Receive hook version, explode here or fail quietly and let the caller figure it out?

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
```diff
@@ -65,6 +66,19 @@ impl FRC46Token<RuntimeError> for BasicToken<'_> {
...
+            ret
+        } else {
+            // state has changed, update return data with new balances
+            self.util.load_replace(&new_cid)?;
+            TransferReturn {
+                from_balance: self.balance_of(operator)?,
+                to_balance: self.balance_of(to)?,
+                recipient_data: ret.recipient_data,
+            }
+        };
```
anorth, line 80: Seeing this, I'm now wondering about yet another change to the hook return to encourage safety. A token impl should generally re-inspect state to create the return value. There's a possible optimisation where the CID hasn't changed, but the general case is to reload.

So now I think we should _not_ return the ret from the hook, but force the caller to recreate it. But make that easy by adding token methods that create a TransferReturn, MintReturn etc from the current state. The hook return would only contain the recipient data.

So, sketching...

```
let hook_ret = hook.call()?;
self.reload()?; // reloads state only if CID has changed
return self.util.transfer_return(hook_ret);
```

testing/fil_token_integration/actors/test_actor/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+pub enum TestAction {
+    /// Accept the tokens
+    Accept,
+    /// Reject the tokens (hook aborts)
+    Reject,
+    /// Transfer to another address (with operator_data that can provide further instructions)
+    Transfer(Address, RawBytes),
+    /// Burn incoming tokens
+    Burn,
+}
```
anorth, line 43: Nice

testing/fil_token_integration/actors/test_actor/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+                if !balance_receipt.exit_code.is_success() {
+                    panic!("unable to get balance");
+                }
+                balance_receipt.return_data.deserialize::<TokenAmount>().unwrap()
+            };
+
+            match params.action {
+                TestAction::Accept => {
+                    // nothing to do here
+                    NO_DATA_BLOCK_ID
```
anorth, line 145: Abort for safety? This means the caller did somethign wrong

testing/fil_token_integration/actors/test_actor/src/lib.rs
anorth: I reckon fail for both

testing/fil_token_integration/tests/common.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn install_actor_stateless(&mut self, path: &str, actor_id: u64) -> Address {
+        let code = load_actor_wasm(path);
+        let address = Address::new_id(actor_id);
+        self.set_actor_from_bin(&code, Cid::default(), address, TokenAmount::zero()).unwrap();
+        address
+    }
+}
+
+impl<B: Blockstore, E: Externs> TokenHelpers for Tester<B, E> {
+    fn get_balance(
```
anorth, line 139: Call it token balance to make slightly less confusing with the native FIL token balance

testing/fil_token_integration/actors/test_actor/src/lib.rs
anorth: I would say it should abort and so should transfer, but I see you're using this return value in tests, so ok to keep this behaviour if it's useful.

testing/fil_token_integration/tests/frc46_multi_actor_tests.rs
```diff
@@ -0,0 +1,171 @@
...
+        let ret_val = tester.call_method(operator[0].1, actor, method_hash!("Constructor"), None);
+        assert!(ret_val.msg_receipt.exit_code.is_success());
+    }
+
+    // TEST: alice sends bob a transfer of zero amount (rejecting first time and then accepting)
+    // first, tell bob to reject it
+    let params = action_params(token_actor, TestAction::Transfer(bob, action(TestAction::Reject)));
+    let ret_val = tester.call_method(operator[0].1, alice, method_hash!("Action"), Some(params));
+    // we told bob to reject, so the action call should return success but give us the error result as return data
+    assert!(ret_val.msg_receipt.exit_code.is_success());
```
anorth, line 51: Suggestion: add tester methods to compress this. E.g. `tester.call_method_ok` which includes asserting the call succeeded

testing/fil_token_integration/tests/frc46_multi_actor_tests.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    // this time tell bob to accept it
+    let params = action_params(token_actor, TestAction::Transfer(bob, action(TestAction::Accept)));
+    let ret_val = tester.call_method(operator[0].1, alice, method_hash!("Action"), Some(params));
+    // we told bob to accept this time, so transfer should succeed
+    assert!(ret_val.msg_receipt.exit_code.is_success());
+
+    // balance should remain zero
+    let balance = tester.get_balance(operator[0].1, token_actor, alice);
+    assert_eq!(balance, TokenAmount::from_atto(0));
```
anorth, line 64: Add a `tester.assert_balance(x)` method, and maybe even `tester.assert_balance_zero()`

testing/fil_token_integration/tests/frc46_multi_actor_tests.rs
```diff
@@ -0,0 +1,171 @@
...
+    let balance = tester.get_balance(operator[0].1, token_actor, bob);
+    assert_eq!(balance, TokenAmount::from_atto(0));
+
+    // TEST: alice sends bob a transfer of a non-zero amounnt. As before, we'll reject it the first time then accept
+    // mint some tokens to alice first
```
anorth, line 69: Suggestion: wrap each independent test in a block to make it more obvious, and avoid accidental variable scope

testing/fil_token_integration/tests/frc46_multi_actor_tests.rs
```diff
@@ -0,0 +1,171 @@
...
+    assert!(ret_val.msg_receipt.exit_code.is_success());
+    let balance = tester.get_balance(operator[0].1, token_actor, alice);
+    assert_eq!(balance, TokenAmount::from_atto(100));
+    // now send to bob, who will reject them
+    let params = action_params(token_actor, TestAction::Transfer(bob, action(TestAction::Reject)));
+    let ret_val = tester.call_method(operator[0].1, alice, method_hash!("Action"), Some(params));
+    assert!(ret_val.msg_receipt.exit_code.is_success());
+    // check the receipt we got in return data
+    let receipt = ret_val.msg_receipt.return_data.deserialize::<Receipt>().unwrap();
+    assert!(!receipt.exit_code.is_success());
```
anorth, line 86: Critical to assert all the balances here, and after every transfer.

testing/fil_token_integration/tests/frc46_multi_actor_tests.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    // TEST: mint to alice who transfers to bob inside receiver hook, bob accepts
+    let ret_val = tester.mint_tokens(
+        operator[0].1,
+        token_actor,
+        alice,
+        TokenAmount::from_atto(100),
+        action(TestAction::Transfer(bob, action(TestAction::Accept))),
+    );
+    assert!(ret_val.msg_receipt.exit_code.is_success());
```
anorth, line 106: Suggestion `tester.mint_ok()` that asserts success

testing/fil_token_integration/tests/frc46_single_actor_tests.rs
```diff
@@ -0,0 +1,146 @@
...
+mod common;
+use common::{construct_tester, TestHelpers, TokenHelpers};
+use test_actor::{action, ActionParams, TestAction};
+
+const BASIC_TOKEN_ACTOR_WASM: &str =
+    "../../target/debug/wbuild/basic_token_actor/basic_token_actor.compact.wasm";
+const TEST_ACTOR_WASM: &str = "../../target/debug/wbuild/test_actor/test_actor.compact.wasm";
+
+/// This covers several simpler tests, which all involve a single receiving actor
+/// They're combined because these integration tests take a long time to build and run
```
anorth, line 17: This is just commentary, but I'm really surprised if breaking this into multiple `#[test]` functions adds any observable build/runtime. What kind of crappy dev environment is this (says Rust n00b)?

testing/fil_token_integration/tests/frc46_single_actor_tests.rs
```diff
@@ -0,0 +1,146 @@
...
+    // Instantiate machine
+    tester.instantiate_machine(DummyExterns).unwrap();
+
+    // construct actors
+    let ret_val = tester.call_method(operator[0].1, token_actor, method_hash!("Constructor"), None);
+    assert!(ret_val.msg_receipt.exit_code.is_success());
+    let ret_val = tester.call_method(operator[0].1, test_actor, method_hash!("Constructor"), None);
+    assert!(ret_val.msg_receipt.exit_code.is_success());
+
+    // TEST: mint to test actor who rejects hook
```
anorth, line 49: Use blocks to distinguish test cases here too

testing/fil_token_integration/tests/frc46_single_actor_tests.rs
anorth: Oh, this is to do with the fvm integration test environment, compiling to WASM and stuff? Ok

frc46_token/src/token/mod.rs
```diff
@@ -105,6 +105,13 @@ where
         self.state = state;
     }
 
+    /// Loads a fresh copy of the state from a blockstore from a given cid, replacing existing state
+    /// The old state is returned to enable comparisons and the like but can be safely dropped otherwise
+    pub fn load_replace(&mut self, cid: &Cid) -> Result<TokenState> {
```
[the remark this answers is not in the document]
abright: > We'll need `replace` to be made properly public too.

I'll change it to use `std::mem::replace` like the `load_replace` function does, that should make it safe for general use. Will also rearrange them so they're grouped together.

testing/fil_token_integration/actors/test_actor/src/lib.rs
abright: Will do

testing/fil_token_integration/tests/common.rs
abright: Will do. I'm also thinking of splitting it off into another module later - we'd have `common.rs` with the basic helpers for calling methods and installing actors but then separate FRC46 and NFT modules for things like minting, balance checks, etc.

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
abright: I think things get simpler with an approach like that. Might want to rename the underlying functions from `mint` to `mint_init` or something to make clearer that it's a multi-step process? I'll leave that alone for the moment but it's a quick enough change afterwards if we want it in there.

The `reload` in your example probably needs to take the original CID we got when flushing the state before the hook call, but this otherwise looks like a simple enough change. Will try it out once I've addressed other review feedback as it's a more complex change than those.

testing/fil_token_integration/tests/frc46_multi_actor_tests.rs
abright: Think I've covered all these now.

Also added some extra checks of `TransferReturn` data in a couple of the more complex cases. I want to do more of that in a follow-up along with the remaining few test cases. The multi-layered stuff gets messy real quick because each struct serializes the next layer in its `return_data` or `recipient_data`. I want to spend some time coming up with a nicer way to handle that stuff but I figure it's a low priority for the moment.

testing/fil_token_integration/actors/basic_token_actor/src/lib.rs
abright: Had a go at it, definitely feels nicer to use when implementing a token.

frc46_token/src/token/mod.rs
```diff
@@ -76,11 +77,24 @@ where
         Self { bs, msg, granularity, state }
     }
 
+    /// Replace the current state with another
+    /// The previous state is returned and can be safely dropped
+    pub fn replace(&mut self, state: TokenState) -> TokenState {
+        std::mem::replace(self.state, state)
```
anorth, line 83: I don't think this is what I want. I think I was confused when I asked for it.

What I need would be be parallel with `wrap`, taking a mutable reference. I have a new `State` (embedded in another struct) that I've loaded into memory, and now want a token wrapping it. I want future mutations to mutate that reference in place (so I can persist later). I don't want to further mutate the state behind the old reference.

Perhaps I could just `wrap` again to get a new token in this case, but a `re-wrap(&mut new_state)` would still be convenient.

We should probably restore the old `replace` method for its use in testing.

frc46_token/src/token/mod.rs
```diff
@@ -159,9 +167,9 @@ where
 
         // Increase the balance of the actor and increase total supply
         let result = self.transaction(|state, bs| {
-            let balance = state.change_balance_by(&bs, owner_id, amount)?;
-            let supply = state.change_supply_by(amount)?;
-            Ok(MintReturn { balance, supply: supply.clone(), recipient_data: RawBytes::default() })
+            state.change_balance_by(&bs, owner_id, amount)?;
+            state.change_supply_by(amount)?;
```
anorth, line 171: Discussion: in most cases, when there is no recursion, the value returned here will be the right one. It would be nice to avoid traversing the balance table again to look up the value when that happens. So storing it in the intermediate for that case could help. That then raises the question of flowing information through to the token so that it knows whether it needs to look up a new balance. Maybe an internal dirty flag or mutation count?

I think this is a nuanced enough issue that you should write an issue and defer it to a new PR. Let's get it correct now, and then optimise.

frc46_token/src/token/types.rs
```diff
@@ -129,7 +129,17 @@ pub struct MintReturn {
 }
 
 impl Cbor for MintReturn {}
-impl RecipientData for MintReturn {
+
+/// Intermediate data used by mint_return to construct the return data
+#[derive(Debug)]
+pub struct MintIntermediate {
+    /// Recipient address to use for querying balance
+    pub recipient: Address,
```
anorth, line 137: Should we restrict visibility of these to the crate? I'm undecided, so ok to leave public for now.

testing/fil_token_integration/actors/test_actor/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+                    let burn_params = BurnParams {
+                        amount: balance,
+                    };
+                    let receipt = sdk::send::send(&params.token_address, method_hash!("Burn"), RawBytes::serialize(&burn_params).unwrap(), TokenAmount::zero()).unwrap();
+                    if !receipt.exit_code.is_success() {
+                        panic!("burn call failed");
+                    }
+                    NO_DATA_BLOCK_ID
+                }
+            }
```
anorth, line 179: Nit: I'd say it's worth pulling this out of `invoke` to a new function now

anorth (approved):

frc46_token/src/token/mod.rs
abright: What you're describing might be better done with `std::mem::swap`. Though it may also be better to just treat the `Token` as disposable and create a new one as needed. Which I guess is what we'd end up doing in a re-wrap function - create a new Token by taking everything but the state from the old one.

This new `replace` doesn't hurt the tests, it's a better version of what they were using before.

frc46_token/src/token/mod.rs
abright: Made issue https://github.com/helix-onchain/filecoin/issues/110 for it

PR #104, opened 2022-09-14
allow hamts to be created with variable bit width

alexytsu:
This impl requires the bit_width to be saved in state but having it piped through from the token level could get quite messy as it's really an internal impl detail of TokenState


anorth (approved):
Storing it in state makes sense. I actually wish the HAMT already did this in its root node.

PR #108, opened 2022-09-16
NFT Mint

alexytsu:
A bit of cleanup for `fvm_actor_utils`

Adds tests and implementation for minting single NFTs


frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+    Other(String),
+}
+
+/// Each token stores its owner, approved operators etc.
+#[derive(Serialize_tuple, Deserialize_tuple, Clone, Debug)]
+pub struct TokenData {
+    pub owner: ActorID,
+    // operators on this token
+    pub approved: Vec<ActorID>, // or maybe as a Cid to an Amt
+    pub metadata_uri: String,
```
anorth, line 34: Assuming the spec will not require URI, use `metadata_key` or `metadata_id` or similar.

frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+pub enum StateError {
+    #[error("ipld amt error: {0}")]
+    IpldAmt(#[from] AmtError),
+    #[error("ipld hamt error: {0}")]
+    IpldHamt(#[from] HamtError),
+    #[error("other error: {0}")]
+    Other(String),
+}
+
+/// Each token stores its owner, approved operators etc.
```
anorth, line 28: It might be worth profiling this vs a less normalised approach after we've done more. This scheme can reduce the number of AMT traversals when fetching all the token data, but requires loading irrelevant bytes from state when only one field of the data is needed (c.f. row vs column stores). The metadata key in this case is likely to dominate the storage costs, but how often will it be read or written or-chain?

This metadata is also probably the thing most likely to be customised by users.

frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Serialize_tuple, Deserialize_tuple, Clone, Debug)]
+pub struct TokenData {
+    pub owner: ActorID,
+    // operators on this token
+    pub approved: Vec<ActorID>, // or maybe as a Cid to an Amt
```
anorth, line 33: `operators`?

frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+            owner_data: empty_owner_map,
+            next_token: 0,
+            total_supply: 0,
+        })
+    }
+
+    pub fn load<BS: Blockstore>(store: &BS, root: &Cid) -> Result<Self> {
+        match store.get_cbor::<Self>(root) {
+            Ok(Some(state)) => Ok(state),
+            _ => panic!(""),
```
anorth, line 85: StateError?

frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+        };
+        let block = Block { codec: DAG_CBOR, data: serialized };
+        let cid = match store.put(Code::Blake2b256, &block) {
+            Ok(cid) => cid,
+            Err(err) => return Err(StateError::Other(err.to_string())),
+        };
+        Ok(cid)
+    }
+
+    fn get_token_data_amt<'bs, BS: Blockstore>(
```
anorth, line 102: Nit: make these public for token users and other usecases

frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+        metadata_uri: String,
+    ) -> Result<TokenID> {
+        // update token data array
+        let mut token_array = self.get_token_data_amt(bs)?;
+        let token_id = self.next_token;
+        token_array.set(token_id, TokenData { owner, approved: vec![], metadata_uri })?;
+
+        // update owner data map
+        let mut owner_map = self.get_owner_data_hamt(bs)?;
+        let new_owner_data = match owner_map.delete(&owner) {
```
anorth, line 132: I think `delete` will do some recomputation of intermediate nodes, which is a waste when we're going to put a value back at the same path.

frcxx_nft/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+            Err(e) => return Err(e.into()),
+        };
+        owner_map.set(owner, new_owner_data)?;
+
+        // update global trackers
+        self.next_token += 1;
+        self.total_supply += 1;
+
+        self.token_data = token_array.flush()?;
+        self.owner_data = owner_map.flush()?;
```
anorth, line 149: Note receiver hook will go here, so might want to prepare for that pattern now.

anorth (approved):

Issue #94, opened 2022-09-06
Extract universal receiver hook out of token library and into its own package


abright:
Follow-up from issue https://github.com/helix-onchain/filecoin/issues/88 and PR https://github.com/helix-onchain/filecoin/pull/93#discussion_r963127402

The receiver hook itself isn't limited to FRC46 tokens anymore and should find use with NFTs and other assets. It's worth pulling that out into a separate crate so it's easier to share around and use without pulling in all the token code.

Issue #77, opened 2022-08-23
Make the FRC interface/trait generic on errors


alexytsu:
https://github.com/helix-onchain/filecoin/pull/75#discussion_r952105866

anorth:
FWIW I've just not used the trait in built-in actors.

anorth:
We make the error an associated type, so I think this is done.

PR #109, opened 2022-09-22
Split Universal Receiver into its own package

alexytsu:
Addresses https://github.com/helix-onchain/filecoin/issues/94

- Moves the universal receiver into its own package and made the code more generic to support potential other future tokens 
- Retained FRC_46 shortcut and convenience methods which enforces that the payload and `token_type` is set as expected

Addresses https://github.com/helix-onchain/filecoin/issues/77 

- Simplifies the error interface a bit for the FRC46Token trait
- Delegates the From<?> to ExitCode into each error's file rather than the monlithic impl


Cargo.toml
```diff
@@ -1,18 +1,18 @@
...
     "frc42_dispatch/hasher",
     "frc42_dispatch/macros",
     "frc42_dispatch/macros/example",
-    "dispatch_examples/greeter",
-    "fvm_dispatch_tools",
     "frc46_token",
     "frcxx_nft",
+    "fvm_actor_utils",
+    "fvm_dispatch_tools",
     "testing/fil_token_integration",
```
alexytsu, line 13: @abright to help make merge conflicts a bit clearer here I've sorted this alphabetically so that if packages are duplicated they'll show up obviously

fvm_actor_utils/src/receiver/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+    #[error("error encoding to ipld")]
+    IpldEncoding(#[from] fvm_ipld_encoding::Error),
+    #[error("error sending message")]
+    Messaging(#[from] MessagingError),
+    #[error("receiver hook error from {address:?} when called with {receiver_params:?}: exit_code={exit_code:?}, return_data={return_data:?}")]
+    Receiver {
+        address: Address,
+        exit_code: ExitCode,
+        return_data: RawBytes,
+        receiver_params: UniversalReceiverParams,
```
alexytsu, line 43: This error replaces the old 
```
 ReceiverHook {
        /// Whose balance is being debited
        from: ActorID,
        /// Whose balance is being credited
        to: ActorID,
        /// Who initiated the transfer of funds
        operator: ActorID,
        amount: TokenAmount,
        exit_code: ExitCode,
    },
``` 

The new error returns more information about the message itself and the receiving actor rather than semantics related to the transfer

Cargo.toml
abright: Nice. I remember using a vscode plugin to do stuff like this in the past, "Prettier TOML" or "Even Better TOML" or something like that.

fvm_actor_utils/src/receiver/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+            address,
+            token_params,
+            token_type,
+            called: false,
+            result_data: Some(result_data),
+        }
+    }
+
+    /// Construct a new FRC46 ReceiverHook call
+    pub fn new_frc46(
```
abright, line 116: If we find that we need more of these helper/shortcut functions, we could implement them in traits instead. So we'd just `use frc46_token::ReceiverHook` (for example) and get frc46-specific helper functions. I don't think it's something we need to do for now but worth keeping in mind as hook usage expands into NFTs and beyond.

fvm_actor_utils/src/receiver/mod.rs
abright: When this error is returned, we should still have the original `ReceiverHook` struct available so I don't think we need the `UniversalReceiverParams`. Might be worth keeping the type from it for error messages, then adding some getter functions to `ReceiverHook` to query other data (or making some fields public maybe).

The important data here would be the `exit_code` and `return_data`, because they're not available elsewhere. Return data might not be useful either if the hook aborts, I don't think we get any return data in that case?

abright:
> @abright I'm slightly confused by the CI failures here.
> 
> I can't replicate it locally - do you have any idea what's going on? Is it a real problem?

It's complaining that the error type is too big, which forces the compiler to reserve and move enough memory to cover it each time regardless of the actual result. See https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

The suggestion there is to box up whatever the big chunk of data is, but it's probably worth taking a look first at what the error type contains to see if there's things we could trim from it.

I reproduced it on the more powerful of my two machines here after `cargo clean` and deleting `Cargo.lock`

fvm_actor_utils/src/receiver/mod.rs
alexytsu: Makes sense - and it seems this was the error in question that was too large so removing it also fixes the CI error

fvm_actor_utils/src/receiver/frc46.rs
```diff
@@ -0,0 +1,27 @@
...
+use fvm_ipld_encoding::{Cbor, RawBytes};
+use fvm_shared::{econ::TokenAmount, ActorID};
+
+use super::ReceiverType;
+
+pub const FRC46_TOKEN_TYPE: ReceiverType = method_hash!("FRC46") as u32;
+
+/// Receive parameters for an FRC46 token
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct FRC46TokenReceived {
```
anorth, line 12: I kinda think this should go in the FRC46 package rather than here, but it's a weak desire, and I haven't traced through to understand the resulting deps. Certainly we shouldn't end up with this crate depending on frc46.

fvm_actor_utils/src/receiver/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+}
+
+/// Parameters for universal receiver
+///
+/// Actual payload varies with asset type
+/// eg: FRC46_TOKEN_TYPE will come with a payload of FRC46TokenReceived
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct UniversalReceiverParams {
```
anorth, line 60: Nit: I'd put this at the top of the file, before its use.

anorth (approved):

fvm_actor_utils/src/receiver/frc46.rs
abright: I think it only needs to be available for the `new_frc46` helper (see https://github.com/helix-onchain/filecoin/pull/109/files#diff-8dc699e4ff6e5f5e0e14a7ddec8ef0414438e44cd18e6eccc1ecaa3ff4f2f6d9R106) but that could be resolved by implementing the function in a separate helper trait inside the `frc46_token` crate. Then token library users would simply `use frc46_token::FRC46Receiver` (or whatever we name it) to have that method available. Might be worth revisiting the idea to keep the token-specific parts separate as the receiver itself is meant to be universal

alexytsu: I've split this out now and will follow the same pattern for NFTs rather than bloating the universal package

PR #113, opened 2022-09-24
fix: balances map should be keyed by bytes

### LINKED ISSUE #112: Spec compliant HAMT keys
The fungible token state uses ActorIDs directly as keys in the HAMTs in its state.  This is not spec compliant because ActorIDs serialize as cbor ints.  This is something that the HAMT library should have caught with type checking but unfortunately this is not yet the case (https://github.com/filecoin-project/ref-fvm/issues/900).  

For now the right way to handle this is to wrap integer serializing keys in a byte serializing type here's [an example in the power actor](https://github.com/filecoin-project/builtin-actors/blob/master/actors/power/src/state.rs#L379).

arajasek:
Fixes #112

### DIFF
--- a/frc46_token/src/token/state.rs
+++ b/frc46_token/src/token/state.rs
@@ -9,12 +9,13 @@ use fvm_ipld_encoding::tuple::*;
 use fvm_ipld_encoding::Cbor;
 use fvm_ipld_encoding::CborStore;
 use fvm_ipld_encoding::DAG_CBOR;
-use fvm_ipld_hamt::Error as HamtError;
 use fvm_ipld_hamt::Hamt;
+use fvm_ipld_hamt::{BytesKey, Error as HamtError};
 use fvm_shared::address::Address;
 use fvm_shared::bigint::Zero;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::ActorID;
+use integer_encoding::VarInt;
 use thiserror::Error;
 
 /// This value has been chosen to optimise to reduce gas-costs when accessing the balances map. Non-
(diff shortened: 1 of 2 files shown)


frc46_token/src/token/state.rs
```diff
@@ -509,18 +511,24 @@ impl TokenState {
         let mut balance_sum = TokenAmount::zero();
         let mut maybe_err: Option<StateInvariantError> = None;
         let balances = self.get_balance_map(bs)?;
-        let res = balances.for_each(|owner, balance| {
+        let res = balances.for_each(|owner_key, balance| {
+            let owner = match u64::decode_var(owner.0.as_slice()) {
```
anorth, line 515: Could you pull this decode out to a nicely-named function alongside actor_id_key?

anorth (approved):

abright:
CI error log suggests there's a dependency missing here, maybe `integer-encoding` like #114 is doing?

frc46_token/src/token/state.rs
arajasek: Done!

arajasek:
@abright Thanks, I've added the dep. I'm not able to run CI myself, and can't check locally because I run into the following failure:

```
error[E0432]: unresolved imports `std::any::Demand`, `std::any::Provider`
 --> /home/aayush/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.36/src/provide.rs:1:16
  |
1 | use std::any::{Demand, Provider};
  |                ^^^^^^  ^^^^^^^^ no `Provider` in `any`
  |                |
  |                no `Demand` in `any`

error[E0635]: unknown feature `provide_any`
   --> /home/aayush/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.36/src/lib.rs:236:34
    |
236 | #![cfg_attr(provide_any, feature(provide_any))]
    |                                  ^^^^^^^^^^^
```

1) Can you please approve the CI action for me?
2) Any ideas about the cause for my failure to build locally? I tried updating / locking down versions of anyhow and thiserror, but it didn't work.

PR #114, opened 2022-09-26
use bytes keys for hamts indexed by actorid

alexytsu:
Equivalent of https://github.com/helix-onchain/filecoin/pull/113 for the NFT library


arajasek:
Hey, sorry, I think #113 was actually wrong. I'm opening a fix soon, and can also include the correct impl for this library.

PR #115, opened 2022-09-26
Update hasher

### LINKED ISSUE #111: Method hash macro uses wrong hash function
The method hash macro uses blake2b-256, which is different than the FVM version (blake2b-512), changed in #89.

When fixing this, please create some "golden" tests for both the FVM and macro versions that include some hardcoded expected method number results that are computed via some separate code written in another programming lanuage.

abright:
Fixes #111 

The hashing function was correct but we were missing the separator tag which has now been added.

I've also added a python implementation of the method number function for generating golden values to use in tests. This gives us a separate implementation to compare against when testing.

The token actor code and the macro test code have both been updated with new method numbers to reflect these changes

### DIFF
--- a/frc42_dispatch/generate_hashes.py
+++ b/frc42_dispatch/generate_hashes.py
@@ -1,0 +1,27 @@
+from hashlib import blake2b
+
+# See https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0042.md#method-number-computation
+def method_number(name):
+    name = '1|' + name
+    hash = blake2b(name.encode('ascii'), digest_size=64)
+    #print('digest: ' + hash.hexdigest())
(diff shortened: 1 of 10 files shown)


frc42_dispatch/generate_hashes.py
```diff
@@ -0,0 +1,27 @@
+from hashlib import blake2b
+
+# See https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0042.md#method-number-computation
+def method_number(name):
+    name = '1|' + name
+    hash = blake2b(name.encode('ascii'))
```
anorth, line 6: Pass `digest_size=64` explicitly just to be concrete (but it is the default too).

anorth (approved):

PR #121, opened 2022-09-27
fix: correctly key ID addresses in maps

arajasek:
Sorry, #113 wasn't how we actually key addresses in Filecoin today, and we should stick to the existing precedent.

The protocol is to serialize the address's payload (including its protocol).


anorth:
No, this was an explicit design decision to key by actor ID instead of address. It locks in the status as definitely being a resolved address. This is a better way. It was discussed (in the context of message parameters) [here](https://github.com/filecoin-project/FIPs/discussions/407#discussioncomment-3312126).

Note that the new HAMTS in the built-in verified registry do the same thing. But I've just realised they probably have the same problem too.

arajasek:
@anorth Thanks for the context -- I'm happy to close this given the decision (which i think is well-motivated).

PR #125, opened 2022-10-03
Fix allowances map keys to be BytesKey

anorth:
This applies the change in #113 to the allowances map too. We were all too hasty making that change.

I've also extracted type aliases that would have made this harder to overlook the first time.

FYI @arajasek


abright (approved):
Looks like the `owner` at https://github.com/helix-onchain/filecoin/blob/anorth/allowance-map-key/frc46_token/src/token/state.rs#L772 needs to be a `BytesKey` now?

LGTM aside from that

Issue #124, opened 2022-10-03
Return a state summary from check_invariants


anorth:
Callers of the state invariant check will often want to check additional app-specific invariants, including reconciliation with the state of other actors.

Return a summary of the state from check_invariants. This doesn't need to be CBOR-serializable.
- TreeMap of all balances
- TreeMap of all allowances
- Total supply

Stebalien:
Also fixed?

Issue #123, opened 2022-10-03
Check granularity of balances in token state invariant check


anorth:
The balance granularity is not stored in state, but supplied from outside. Check that all balances are whole multiples of the granularity.

PR #126, opened 2022-10-04
Improvements to check_invariants

alexytsu:
Addresses: 
- https://github.com/helix-onchain/filecoin/issues/123
- https://github.com/helix-onchain/filecoin/issues/124

### DIFF
--- a/dispatch_examples/greeter/Cargo.toml
+++ b/dispatch_examples/greeter/Cargo.toml
@@ -14,7 +14,7 @@ fvm_shared = { version = "2.0.0-alpha.2" }
 cid = { version = "0.8.5", default-features = false }
 fvm = { version = "2.0.0-alpha.2", default-features = false }
 fvm_integration_tests = "2.0.0-alpha.1"
-actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", branch = "next", features = ["m2-native"] }
+actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", rev = "6d800c449c0a59bbc4728a68924b4c64fbff130d", features = ["m2-native"] }
 
 [build-dependencies]
 wasm-builder = "3.0"
--- a/dispatch_examples/greeter/Cargo.toml
+++ b/dispatch_examples/greeter/Cargo.toml
@@ -14,7 +14,7 @@ fvm_shared = { version = "2.0.0-alpha.2" }
 cid = { version = "0.8.5", default-features = false }
 fvm = { version = "2.0.0-alpha.2", default-features = false }
 fvm_integration_tests = "2.0.0-alpha.1"
-actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", rev = "6d800c449c0a59bbc4728a68924b4c64fbff130d", features = ["m2-native"] }
+actors-v9 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", rev = "b40284a166adb22612b4fd98ab6c2483f581d8f7" }
 
 [build-dependencies]
 wasm-builder = "3.0"
--- a/frc46_token/src/token/state.rs
+++ b/frc46_token/src/token/state.rs
@@ -54,8 +54,6 @@ pub enum StateError {
 
 #[derive(Error, Debug)]
 pub enum StateInvariantError {
-    #[error("ipld hamt error: {0}")]
-    IpldHamt(#[from] HamtError),
     #[error("total supply was negative: {0}")]
     SupplyNegative(TokenAmount),
     #[error("the account for {account:?} had a negative balance of {balance:?}")]
@@ -161,11 +159,7 @@ impl TokenState {
     }
 
     /// Get the balance of an ActorID from the currently stored state
-    pub fn get_balance<BS: Blockstore>(
-        &self,
-        bs: &BS,
-        owner: ActorID,
-    ) -> std::result::Result<TokenAmount, HamtError> {
+    pub fn get_balance<BS: Blockstore>(&self, bs: &BS, owner: ActorID) -> Result<TokenAmount> {
         let balances = self.get_balance_map(bs)?;
 
         let balance = match balances.get(&actor_id_key(owner))? {
@@ -188,7 +182,7 @@ impl TokenState {
     ) -> Result<TokenAmount> {
         if delta.is_zero() {
             // This is a no-op as far as mutating state
-            return Ok(self.get_balance(bs, owner)?);
+            return self.get_balance(bs, owner);
         }
 
         let mut balance_map = self.get_balance_map(bs)?;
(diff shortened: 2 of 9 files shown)


frc46_token/src/token/state.rs
```diff
@@ -523,7 +530,8 @@ impl TokenState {
         let res = balances.for_each(|owner_key, balance| {
             let owner = match decode_actor_id(owner_key) {
                 None => {
-                    bail!("invalid key in balances map")
+                    maybe_err = Some(StateInvariantError::InvalidOwnerKey(owner_key.clone()));
+                    bail!("invariant failed");
```
anorth, line 534: I should have caught this earlier, let's fix this while we're here.

The general idea behind these checks in the builtin actors is to always proceed, but accumulate errors to be returned. The methods in builtin actors don't even return a Result (but they do unwrap() some impossible-to-recover things). They basically return a list of strings describing any problems encountered, and the state summary. [Example](https://github.com/filecoin-project/builtin-actors/blob/master/actors/power/src/testing.rs#L36-L40).

Could we update to a similar pattern here? You can't depend on the MessageAccumulator type, but just returning a vector of strings will be fine.

Also then avoid using `anyhow` for `bail`.

frc46_token/src/token/state.rs
```diff
@@ -627,6 +651,16 @@ pub fn decode_actor_id(key: &BytesKey) -> Option<ActorID> {
 
 impl Cbor for TokenState {}
 
+/// A summary of the current state to allow checking application specific invariants
+pub struct StateSummary<'bs, BS>
+where
+    BS: Blockstore,
+{
+    pub balance_map: BalanceMap<'bs, BS>,
+    pub allowance_map: AllowanceMap<'bs, BS>,
```
anorth, line 660: Please fully load this state from the store so that StateSummary is self-contained and has no generics. I.e. build HashMaps here.

I have [WIP here](https://github.com/filecoin-project/builtin-actors/blob/cae57d2d67e529eeed3f51f53ab79a8e08afa22f/actors/datacap/src/testing.rs#L12-L16) building these for some testing I need, but I expect to just import this version in the future, whereupon my predicates can just use the state summary directly.

anorth (changes requested):

dispatch_examples/greeter/Cargo.toml
```diff
@@ -14,7 +14,7 @@ fvm_shared = { version = "2.0.0-alpha.2" }
 cid = { version = "0.8.5", default-features = false }
 fvm = { version = "2.0.0-alpha.2", default-features = false }
 fvm_integration_tests = "2.0.0-alpha.1"
-actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", branch = "next", features = ["m2-native"] }
+actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", rev = "6d800c449c0a59bbc4728a68924b4c64fbff130d", features = ["m2-native"] }
```
alexytsu, line 17: this fixes the CI errors for now. I'll delay upgrading `fvm` `fvm_shared` `fvm_integration_tests` etc. until there's a published crate for v10 of builtin-actors

dispatch_examples/greeter/Cargo.toml
```diff
@@ -14,7 +14,7 @@ fvm_shared = { version = "2.0.0-alpha.2" }
 cid = { version = "0.8.5", default-features = false }
 fvm = { version = "2.0.0-alpha.2", default-features = false }
 fvm_integration_tests = "2.0.0-alpha.1"
-actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", branch = "next", features = ["m2-native"] }
```
anorth, line 17: `next` wasn't the right branch anyway, you probably want to be on master. Where is this commit from?

Feel free to change to a recent master commit.

frc46_token/src/token/state.rs
```diff
@@ -506,114 +511,169 @@ impl TokenState {
...
-    /// where operator == owner.
+    /// where operator == owner. Checks that all balances are a multiple of the granularity.
+    ///
+    /// Returns a state summary that can be used to check application specific invariants.
     pub fn check_invariants<BS: Blockstore>(
         &self,
         bs: &BS,
-    ) -> std::result::Result<(), StateInvariantError> {
+        granularity: u64,
+    ) -> std::result::Result<StateSummary, Vec<StateInvariantError>> {
```
anorth, line 521: Can you drop the Result and just return a tuple? We still want the state summary even if there are problems. Goal is to find all the errors in the whole system in a single run (though we can't get all the way there of course).

frc46_token/src/token/state.rs
```diff
@@ -506,114 +511,169 @@ impl TokenState {
...
-            }
-            // zero balances should not be stored in the Hamt
-            if balance.is_zero() {
-                maybe_err = Some(StateInvariantError::ExplicitZeroBalance(owner));
-                bail!("invariant failed")
+        let balances = match self.get_balance_map(bs) {
+            Ok(map) => map,
+            Err(e) => {
+                errors.push(StateInvariantError::State(e));
+                return Err(errors);
```
anorth, line 548: Move the body of the balance iteration inside the Ok branch so that you can continue checking other parts of the state even if balances can't be loaded.

frc46_token/src/token/state.rs
```diff
@@ -506,114 +511,169 @@ impl TokenState {
...
+                            granularity,
+                        });
+                    }
+
+                    // track total balance
+                    balance_sum = balance_sum.clone() + balance.clone();
+
+                    // clone into HashMap
+                    balance_map.insert(owner, balance.clone());
+                }
```
anorth, line 584: else error about bad key

frc46_token/src/token/state.rs
```diff
@@ -506,114 +511,169 @@ impl TokenState {
...
-                        bail!("invalid operator key in allowances map")
+        let mut global_allowances_map: HashMap<ActorID, BTreeMap<ActorID, TokenAmount>> =
+            HashMap::new();
+        allowances_hamt
+            .for_each(|owner, _| {
+                if let Some(owner) = decode_key_addr(owner, &mut errors) {
+                    let allowance_map = self.get_owner_allowance_map(bs, owner)?;
+
+                    // check that the allowance map exists
+                    if allowance_map.is_none() {
```
anorth, line 614: I don't think this can happen, since we're in for_each. Would be more obvious if you used the cid value rather than doing a get() for it again.

frc46_token/src/token/state.rs
```diff
@@ -627,14 +687,22 @@ pub fn decode_actor_id(key: &BytesKey) -> Option<ActorID> {
 
 impl Cbor for TokenState {}
 
+/// A summary of the current state to allow checking application specific invariants
+#[derive(Clone, Debug)]
+pub struct StateSummary {
+    pub balance_map: HashMap<ActorID, TokenAmount>,
+    pub allowance_map: HashMap<ActorID, BTreeMap<ActorID, TokenAmount>>,
```
anorth, line 694:
```suggestion
    pub allowance_map: HashMap<ActorID, HashMap<ActorID, TokenAmount>>,
```
You copied this from mistaken code that I wrote

anorth:
Nearly there, thanks for bearing with me here.

frc46_token/src/token/state.rs
alexytsu: The intent here is to verify that the allowance map can be looked up by the owner correctly (i.e. that the cid in the global map is not invalid or stale). I've added an explicit check that the cids should match.

anorth (approved):

PR #130, opened 2022-10-07
NFT Transfers

alexytsu:
Some preliminary work for NFT transfers

- [x] `Approve`/`Revoke`, `ApproveForAll`/`RevokeForAll`
- [x] `Transfer`, `TransferFrom`
- [x] Unit tests
- [ ] `IsApproved`/`IsApprovedForAll`
- [ ] Actor level tests
- [X] Single API -> Batching API https://github.com/helix-onchain/filecoin/pull/133
- [ ] Receiver hook handling https://github.com/helix-onchain/filecoin/pull/135

### DIFF
--- a/frcxx_nft/src/state.rs
+++ b/frcxx_nft/src/state.rs
@@ -398,4 +398,7 @@ mod test {
         assert_eq!(state.get_balance(bs, ALICE_ID).unwrap(), 3);
         assert_eq!(state.get_balance(bs, BOB_ID).unwrap(), 0);
     }
+
+    #[test]
+    fn it_allows_delegated_transfers() {}
 }
--- a/frcxx_nft/src/state.rs
+++ b/frcxx_nft/src/state.rs
@@ -1,6 +1,4 @@
 //! Abstraction of the on-chain state related to NFT accounting
-use std::collections::HashSet;
-
 use cid::multihash::Code;
 use cid::Cid;
 use fvm_ipld_amt::Amt;
@@ -17,6 +15,8 @@ use fvm_shared::ActorID;
 use integer_encoding::VarInt;
 use thiserror::Error;
 
+use crate::util::OperatorSet;
+
 pub type TokenID = u64;
 
 /// Each token stores its owner, approved operators etc.
@@ -24,7 +24,7 @@ pub type TokenID = u64;
 pub struct TokenData {
     pub owner: ActorID,
     // operators on this token
-    pub operators: HashSet<ActorID>, // or maybe as a Cid to an Amt
+    pub operators: Vec<ActorID>, // or maybe as a Cid to an Amt
     pub metadata_id: Cid,
 }
 
@@ -33,7 +33,7 @@ pub struct TokenData {
 pub struct OwnerData {
     pub balance: u64,
     // account-level operators
-    pub operators: HashSet<ActorID>, // maybe as a Cid to an Amt
+    pub operators: Vec<ActorID>, // maybe as a Cid to an Amt
 }
 
 /// NFT state IPLD structure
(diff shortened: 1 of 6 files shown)


alexytsu:
I'll defer higher level tests (library-level/actor-level) to a separate PR as it'll force me to deal with receiver hooks as a logically separate change. 

I'd also like to change the current APIs to be batched before making those changes.

I'll target NFT changes into a non-main branch for now whilst there are still todos

alexytsu:
@abright could you please review #133 and this PR when you have a chance?

low urgency as I'm just targeting `nft-dev` anyway. when its a bit more fleshed out, I'll make a PR for `nft-dev` into `main`

frcxx_nft/src/state.rs
```diff
@@ -1,4 +1,6 @@
 //! Abstraction of the on-chain state related to NFT accounting
+use std::collections::HashSet;
```
ZenGround0, line 2: You might already have this worked out but how is the cbor serialization of this type defined and what tradeoffs are motivating using this?  It's showing up on chain below and there's not a lot of precedent for having map datastructures serialized inline in state.  I am suspicious this is going to be strictly worse, in terms of storage overhead and gas cost of operations, than other datatypes more commonly used.

My guess is you either want a HAMT, an inline cbor array, or an inline map type with a more specialized serialization than what you get by sending the standard hash set through cbor tuple encoding.

frcxx_nft/src/state.rs
```diff
@@ -171,6 +180,92 @@ impl NFTState {
...
+            None => OwnerData { balance: 0, operators: HashSet::new() },
+        };
+        owner_map.set(actor_id_key(owner), new_owner_data)?;
+        self.owner_data = owner_map.flush()?;
+
+        Ok(())
+    }
+
+    /// Revokes an operator's premission to transfer tokens on behalf of the owner
+    pub fn revoke_for_all<BS: Blockstore>(
```
ZenGround0, line 246: While `approve_for_all` fits nicely with the logic structure this name doesn't quite match the semantics I would expect given that you can call revoke_for_all for an operator but still have an approved operator transfer [if the token itself does not revoke that operator explicitly](https://github.com/helix-onchain/filecoin/pull/130/files#diff-d366e93293ca5febc8827cd9c7c8c52c4335d6b4b90e79e3c5f50ec8baa025d9R396).

frcxx_nft/src/state.rs
```diff
@@ -17,25 +19,12 @@ use thiserror::Error;
...
-}
-
 /// Each token stores its owner, approved operators etc.
 #[derive(Serialize_tuple, Deserialize_tuple, Clone, Debug)]
 pub struct TokenData {
     pub owner: ActorID,
     // operators on this token
-    pub operators: Vec<ActorID>, // or maybe as a Cid to an Amt
+    pub operators: HashSet<ActorID>, // or maybe as a Cid to an Amt
     pub metadata_id: Cid,
```
ZenGround0, line 28: I'm new to NFT standards so I'm guessing I'm the one who is confused, but I would expect this to be `data_id`.

alexytsu: Are you referring to `pub owner: ActorID`?

`TokenData` is used in `Amt<TokenData>` where the index of the Amt is implicitly the TokenID 

So we need to store the owner, operators, metadata of the NFT in this structure

frcxx_nft/src/state.rs
alexytsu: I suspect you are right. The reference impl will attempt to target the majority use case where I expect the operators list to be comparatively small. So for now, [I've gone with a Vec<ActorID>](https://github.com/helix-onchain/filecoin/pull/130/commits/372b8995bc0f270b467962fd390a500f61d692bf) with order maintained in-memory when operators are added and removed

I'll benchmark this properly against some alternatives when the library is finished to compare options for optimizing gas vs storage etc.

frcxx_nft/src/state.rs
alexytsu: I've renamed it to `approve_for_owner` now but may change it again depending on discussion in the [FIP](https://github.com/filecoin-project/FIPs/discussions/492)

Ideally it'd just match the actor-interface method name

frcxx_nft/src/state.rs
ZenGround0: ah sorry no I'm referring to `metadata_id`. My naive understanding is that NFTs can include arbitrary data which this field accounts for and it is a bit strange to me that it's called metadata given that I think of the data as the main event.  However like I said this is all new to me so feel free to ignore.

PR #133, opened 2022-10-11
singular to batched state apis

alexytsu:
Makes changes to have the APIs be batched

There's probably some more incremental improvements so that multiple state functions can be batched together without flushing the AMTs/HAMTs multiple times. 

e.g. in the use case where multiple tokens are minted then approved for an operator at once, it would be nice for the `Mint` to simply take in &mut to the HAMTs and AMTs and have the caller flush at the end of the entire transaction.

### DIFF
--- a/frcxx_nft/src/state.rs
+++ b/frcxx_nft/src/state.rs
@@ -180,21 +180,25 @@ impl NFTState {
         Ok(balance)
     }
 
-    /// Approves an operator to transfer a specific token
+    /// Approves an operator to transfer a set of specified tokens
     ///
-    /// Checks that the caller is the owner of the specified token
-    pub fn approve_for_token<BS: Blockstore>(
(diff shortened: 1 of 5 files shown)


frcxx_nft/src/state.rs
```diff
@@ -316,20 +348,29 @@ impl NFTState {
...
+        self.token_data = token_array.flush()?;
+        self.owner_data = owner_map.flush()?;
+
+        Ok(())
     }
 
-    /// Makes a transfer of a token from one address to another. The caller must verify that such a transfer is allowed.
+    /// Makes a transfer of a token from one address to another. The caller must verify that such a
+    /// transfer is allowed. This function does not flush the token AMT or the owner HAMT, it is the
+    /// caller's responsibility to do so at the end of the batch.
```
ZenGround0, line 373: 👍

ZenGround0 (approved):
LGTM

PR #135, opened 2022-10-11
NFT Receiver Hooks

alexytsu:


testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
```diff
@@ -55,6 +62,60 @@ fn invoke(params: u32) -> u32 {
...
+            let params = deserialize_params::<TransferParams>(params);
+            handle.transfer(
+                &caller_address(),
+                &params.to,
+                &params.token_ids,
+                params.operator_data,
+                RawBytes::default()
+            ).unwrap();
+            let cid = handle.flush().unwrap();
+            sdk::sself::set_root(&cid).unwrap();
```
abright, line 103: If this calls a receiver hook and flushes state/sets root before that, it won't need to do so here. If it's not doing that, it needs to, otherwise the the receiver hook (which could make further calls to the token actor) will only have the previous state to work with, which will have no record of the transfer.

Looks like minting is doing the right thing already and transfers will follow?

testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
```diff
@@ -55,6 +62,60 @@ fn invoke(params: u32) -> u32 {
...
+            let params = deserialize_params::<TransferFromParams>(params);
+            handle.transfer_from(
+                &caller_address(),
+                &params.to,
+                &params.token_ids,
+                params.operator_data,
+                RawBytes::default()
+            ).unwrap();
+            let cid = handle.flush().unwrap();
+            sdk::sself::set_root(&cid).unwrap();
```
abright, line 116: Same as for Transfer

testing/fil_token_integration/actors/basic_nft_actor/src/lib.rs
alexytsu: Yeh, I'm testing the right structure with minting first and will do Transfer/TransferFrom in a followup

abright (approved):
LGTM, if you want to follow up with more receiver hook stuff in a new branch/PR

PR #138, opened 2022-10-21
generate return value from current state

alexytsu:
Finishes up receiver hooks for transfer and transfer_from

### DIFF
--- a/frcxx_nft/src/lib.rs
+++ b/frcxx_nft/src/lib.rs
@@ -8,6 +8,7 @@
 
 use cid::Cid;
 use fvm_actor_utils::{
+    actor::{Actor, ActorError},
     messaging::{Messaging, MessagingError},
     receiver::ReceiverHook,
 };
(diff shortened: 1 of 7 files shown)


frcxx_nft/src/lib.rs
```diff
@@ -97,6 +114,14 @@ where
         )?)
     }
 
+    /// Constructs MintReturn data from a MintIntermediate handle
+    ///
+    /// Creates an up-to-date view of the actor state where necessary to generate the values
+    pub fn mint_return(&mut self, intermediate: MintIntermediate, cid: Cid) -> Result<MintReturn> {
+        self.reload_if_changed(cid)?;
```
abright, line 121: I don't know if it's a good idea for the library code to be doing this instead of handling state loading on the user implementation side (like we're doing for fungible tokens). It'll be ok for now since it's going into a dev branch but I think it'll cause problems in actor code that carries additional state of its own beyond the `NFTState`

It does open up one easy optimisation though: we can pass through the original `MintReturn` data in the intermediate struct and simply return that if there's no change once the hook call completes. I've been playing with the idea [here](https://github.com/helix-onchain/filecoin/commit/b0a291b700152d955ae33dc7628fa39925e0a00b) (sending original data) and [here](https://github.com/helix-onchain/filecoin/commit/27a35220b174d75f03aa20f2aff042a502a9f521) (trying to centralise the 'has state changed' check) for the fungible token

abright (approved):
LGTM

I think we're going to need to revisit some of the state management/reloading stuff later but it's probably best to do that when the dev branch is ready to merge

frcxx_nft/src/lib.rs
alexytsu: Will keep an eye on this pattern. 

It feels like making sure state is reloaded is a safety/correctness concern that should be enforced by the library. But agree that it limits opportunities for optimisation (especially if there's other state to take care of)

PR #141, opened 2022-10-27
Multi actor integration tests

alexytsu:
This is the first set of integration testing actor-actor functionality for NFTs

Also covers off some functionality that was previously missing from the library e.g. `Burn` `BurnFor` and `BalanceOf`


abright (approved):
LGTM

PR #144, opened 2022-10-29
fix: use upstream wasm-builder

Stebalien:
The 'wasm-builder' crate is a downstream fork. Upstream is 'substrate-wasm-builder'.


alexytsu (approved):
Thanks @Stebalien!

Issue #128, opened 2022-10-05
On chain factory for token contracts


alexytsu:
To investigate the possibility of having a single code bundle on-chain that can be constructed with different params. 

Simplest would be to have an initial account which is permitted to mint that can then be permanently revoked. Or simply a fixed supply to begin with distributed to specified account(s).

anorth:
Everything that would be hard-coded in a token to make it special needs to be in state instead. E.g. name, code etc. And yes configurable minting policy.

Issue #127, opened 2022-10-05
One-line helper to delegate all standardised FRC46 methods


alexytsu:
All standardised method numbers/invocations should be handled by a single line of library code. 

Unhandled method numbers are delegated to the actor author so that they can extend functionality (e.g. for minting)

PR #149, opened 2022-11-03
FRC46 factory token and token invoke() helper

abright:
Implements a token suitable for #128 where the code can be deployed once and multiple tokens created by changing parameters passed into the `Constructor` call. The token name, symbol, and granularity are set at construction time.

Minting is controlled by a single authorised user which is also set at construction time - this address is optional because it's going to be paired with an initial supply (so one or the other must be set and will be checked during construction) to cover the common faucet and fixed supply strategies. A new method `DisableMint` is added that will clear the authorised user to disable minting.

Also introduces `frc46_invoke` which is a generic function to handle dispatch of all standard FRC-0046 token methods (see #127)
This invoke helper function should move out of the token implementation into a library, but it needs a bit more work around error handling so I'll follow up on that later in a new PR.


token_examples/factory_token/src/token.rs
```diff
@@ -0,0 +1,423 @@
...
+        // TODO: token() wants mutable ref, we don't have one here and can't change the interface
+        // so need an immutable wrapper to call these things?
+        // or just bypass it and go to the state directly
+        //self.token().total_supply()
+        self.token.supply.clone()
+    }
+
+    fn balance_of(&self, params: Address) -> Result<BalanceReturn, RuntimeError> {
+        // TODO: same situation as total_supply
+        //Ok(self.token().balance_of(&params)?)
```
abright, line 112: Probably need a separate read-only implementation of the `Token` handle to clean this up (also for `total_supply` and any future query/getter functions)

token_examples/factory_token/src/token.rs
```diff
@@ -0,0 +1,423 @@
...
+        let ret = self.token().mint_return(hook_ret)?;
+
+        Ok(ret)
+    }
+
+    /// Permanently disable minting
+    /// Only the authorised mint operator can do this
+    ///
+    /// TODO: should authorised minter be separate to a token 'owner' ?
+    /// also, should it be a single address, a list of addresses, or some other means?
```
abright, line 308: Thoughts? I did what was easy (the single operator for minting) but not sure if it's the best approach

anorth (commented):
Thanks, this looks directionally right. Biggest thing is to fix the APIs to make things like the token trait and library work together, and enable the generic dispatch thing. 

I didn't look at the tests too much  yet. Is there room to reduce testing overheads by using this actor for the main library integration tests, or is the test actor there too specialised? We do need to refactor to reduce the duplicated code tho.

token_examples/factory_token/src/token.rs
```diff
@@ -0,0 +1,419 @@
...
+        self.symbol.clone()
+    }
+
+    fn granularity(&self) -> GranularityReturn {
+        //1
+        self.granularity
+    }
+
+    fn total_supply(&self) -> TotalSupplyReturn {
+        // TODO: token() wants mutable ref, we don't have one here and can't change the interface
```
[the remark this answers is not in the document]
abright: Generic invoke needs the trait, but maybe it's worth pulling out of where it lives now, reworking it, and keeping it together with the invoke function in some other place - I feel like that stuff belongs in a utility module or library separate to the core token library code. 

I've made the error an associated type which is a bit nicer to live with and doesn't affect existing code too much, still need to do some work on the read-only calls like `total_supply()` though. I'm leaning towards just making them take `&mut self` like most of the other functions already do

token_examples/factory_token/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+use token::{deserialize_params, frc46_invoke, return_ipld, BasicToken, MintParams};
+
+#[derive(Serialize_tuple, Deserialize_tuple, Debug)]
+pub struct ConstructorParams {
+    pub name: String,
+    pub symbol: String,
+    pub granularity: u64,
+    /// authorised mint operator
+    /// only this address can mint tokens or remove themselves to permanently disable minting
+    pub minter: Option<Address>,
```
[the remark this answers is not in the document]
abright: It was a leftover from a plan to have a minter and/or an initial supply amount. We dropped the initial supply part of it so I've removed the option from constructor params now too.

token_examples/factory_token/src/token.rs
abright: It's now a single `ActorID`, as suggested in another comment

token_examples/factory_token/src/token.rs
```diff
@@ -0,0 +1,419 @@
...
+/// TODO: improve the error type/handling so this can be moved into a library and work with any token code 
+///
+pub fn frc46_invoke<T, F>(
+    method_num: u64,
+    params: u32,
+    token: &mut T,
+    flush_state: F,
+) -> Result<Option<u32>, RuntimeError>
+where
+    T: FRC46Token<RuntimeError>,
```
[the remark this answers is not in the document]
abright: Done. I made the return data serialization abort on errors to avoid the minefield of it being compatible with whatever error type a user implements. It should be in a state where it could be used in external code now.

token_examples/factory_token/src/token.rs
```diff
@@ -0,0 +1,419 @@
...
+    #[error("minting has been permanently disabled")]
+    MintingDisabled,
+}
+
+pub fn caller_address() -> Address {
+    let caller = sdk::message::caller();
+    Address::new_id(caller)
+}
+
+// NOTE: this is mostly lifted from the basic token actor we use in integration testing
```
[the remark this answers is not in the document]
abright: Took a bit of crate trickery to set up but the main battery of integration tests now runs and passes using this token

token_examples/factory_token/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    // we only implement our own methods in this handler, anything we don't handle directly is delegated to frc46_invoke
+    // which handles any methods in the FRC46 token interface to save us writing the same
+    match_method!(method_num, {
+        "Constructor" => {
+            let params = deserialize_params(params);
+            construct_token(params);
+            NO_DATA_BLOCK_ID
+        }
+        "Mint" => {
+            let root_cid = fvm_sdk::sself::root().unwrap();
```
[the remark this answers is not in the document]
abright: I've changed it to have invoke() abort on error results and moved the method dispatching into a new function so it can return results. Clippy is mad about the size of the error type but that doesn't seem to be related to changes in this PR - I see it happening for https://github.com/helix-onchain/filecoin/pull/166 as well. I'll address that in a separate branch.

token_examples/factory_token/src/token.rs
abright: Addressed in https://github.com/helix-onchain/filecoin/pull/149#discussion_r1028709951 - I've changed the underlying trait to take mutable refs for these functions

abright:
This now passes our integration tests when run instead of the basic_token_actor. I think we can remove that older token actor now, or maybe move it elsewhere to keep as an example?

I should probably move this one too - drop the `token_example` directory and make it a top level thing, maybe with a different name like `frc46_token_factory` ?

token_examples/factory_token/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+pub fn invoke(params: u32) -> u32 {
+    std::panic::set_hook(Box::new(|info| {
+        fvm_sdk::vm::abort(ExitCode::USR_ASSERTION_FAILED.value(), Some(&format!("{}", info)))
+    }));
+
+    let method_num = fvm_sdk::message::method_number();
+    match token_invoke(method_num, params) {
+        Ok(ret) => ret,
+        Err(err) => {
+            fvm_sdk::vm::abort(ExitCode::USR_ASSERTION_FAILED.value(), Some(&err.to_string()))
```
anorth, line 72: It looks like this will overwrite any exit code provided by the token library when frc46_invoke returns an Err, e.g. insufficient balance

token_examples/factory_token/tests/common.rs
```diff
@@ -0,0 +1,245 @@
...
+        from: Address,
+        to: Address,
+        method_num: u64,
+        params: Option<RawBytes>,
+    ) -> ApplyRet {
+        static mut SEQUENCE: u64 = 0u64;
+        let message = Message {
+            from,
+            to,
+            gas_limit: 99999999,
```
anorth, line 141: This is actually not that high. Can you import BLOCK_GAS_LIMIT to here from fvm_shared?

token_examples/factory_token/tests/factory_token.rs
```diff
@@ -0,0 +1,191 @@
...
+        // try minting some tokens, which should fail
+        let ret_val = tester.mint_tokens(
+            operator[0].1,
+            token_actor,
+            alice,
+            TokenAmount::from_atto(100),
+            action(TestAction::Accept),
+        );
+        assert!(
+            !ret_val.msg_receipt.exit_code.is_success(),
```
anorth, line 183: Please assert the actual exit code expected

anorth (commented):
We might want to rename the factory_token to frc46_factory_token, since there'll probably be an NFT one later too

anorth (approved):
Looks pretty good, but please get review from @alexytsu before merging, especially to discuss traits, dependencies etc.

token_examples/factory_token/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+            // no return
+            Ok(NO_DATA_BLOCK_ID)
+        }
+        _ => {
+            let root_cid = fvm_sdk::sself::root()?;
+
+            let mut token_actor = FactoryToken::load(&root_cid)?;
+
+            // call FRC46 token methods
+            // note that the `token_actor` passed in here needs to know how to save and load state
```
alexytsu, line 39: I'm not quite sure what this means or why it's necessary

token_examples/factory_token/token_impl/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+cid = { version = "0.8.5", default-features = false }
+frc42_dispatch = { path = "../../../frc42_dispatch" }
+frc46_token = { path = "../../../frc46_token" }
+fvm_actor_utils = { version = "0.1.0", path = "../../../fvm_actor_utils" }
+fvm_ipld_blockstore = "0.1.1"
+fvm_ipld_encoding = "0.2.2"
+fvm_sdk = { version = "2.0.0-alpha.2" }
+fvm_shared = { version = "2.0.0-alpha.2" }
```
alexytsu, line 15: as noted above - 2.0.0 is stable

alexytsu (approved):

token_examples/factory_token/src/lib.rs
abright: Added exit codes to the `RuntimeError` type so it can return the proper code here

token_examples/factory_token/tests/factory_token.rs
abright: Done, now that it does return a meaningful exit code for each error

token_examples/factory_token/tests/common.rs
abright: Did this for the main integration tests too

PR #164, opened 2022-11-22
Remove logic from state-level

alexytsu:
The state level beomes a dumb helper that operates on AMT/HAMT data structures

Transaction atomicity and consistency must be maintained at the library level. The library level must check that callers and addresses are valid for the given operations.

### DIFF
--- a/frcxx_nft/src/lib.rs
+++ b/frcxx_nft/src/lib.rs
@@ -15,6 +15,7 @@ use fvm_actor_utils::{
 use fvm_ipld_blockstore::Blockstore;
 use fvm_ipld_encoding::RawBytes;
 use fvm_shared::{address::Address, ActorID};
+use receiver::{FRCXXReceiverHook, FRCXXTokenReceived};
 use state::StateError;
 use thiserror::Error;
 use types::{
@@ -257,19 +258,41 @@ where
         token_data: RawBytes,
     ) -> Result<ReceiverHook<TransferIntermediate>> {
         // Attempt to instantiate the accounts if they don't exist
-        let owner = self.msg.resolve_or_init(owner)?;
-        let recipient = self.msg.resolve_or_init(recipient)?;
+        let owner_id = self.msg.resolve_or_init(owner)?;
+        let recipient_id = self.msg.resolve_or_init(recipient)?;
 
-        let hook = self.state.transfer_tokens(
-            &self.bs,
-            owner,
-            recipient,
-            token_ids,
+        self.transaction(|state, store| {
+            let mut token_array = state.get_token_data_amt(store)?;
+            let mut owner_map = state.get_owner_data_hamt(store)?;
+
+            for token_id in token_ids {
+                let _token_data = NFTState::owns_token(&token_array, owner_id, *token_id)?;
+                // update the token_data to reflect the new owner and clear approved operators
+                state.make_transfer(&mut token_array, &mut owner_map, *token_id, recipient_id)?;
+            }
+
+            state.token_data = token_array.flush().map_err(StateError::from)?;
+            state.owner_data = owner_map.flush().map_err(StateError::from)?;
+
+            Ok(())
+        })?;
+
+        let params = FRCXXTokenReceived {
+            to: recipient_id,
+            operator: owner_id,
+            token_ids: token_ids.into(),
             operator_data,
             token_data,
-        )?;
+        };
 
-        Ok(hook)
+        let intermediate = TransferIntermediate {
+            to: recipient_id,
+            from: owner_id,
+            token_ids: token_ids.into(),
+            recipient_data: RawBytes::default(),
+        };
+
+        Ok(ReceiverHook::new_frcxx(*recipient, params, intermediate).map_err(StateError::from)?)
     }
 
     /// Constructs TransferReturn data from a TransferIntermediate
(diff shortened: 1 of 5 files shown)


frcxx_nft/src/lib.rs
```diff
@@ -179,7 +250,16 @@ where
         let caller = self.msg.resolve_id(caller)?;
         let operator = self.msg.resolve_or_init(operator)?;
 
-        self.state.approve_for_tokens(&self.bs, caller, operator, token_ids)?;
+        self.transaction(|state, bs| {
+            let mut token_array = state.get_token_data_amt(bs)?;
+            NFTState::assert_owns_tokens(&token_array, caller, token_ids)?;
+
+            state.approve_for_tokens(&mut token_array, operator, token_ids)?;
```
anorth, line 257: I think you can use a predicate over token data here too, to avoid looking it up twice. Similar for some other methods below – I won't comment them all.

frcxx_nft/src/lib.rs
```diff
@@ -227,8 +321,14 @@ where
...
         };
 
-        self.state.revoke_for_all(&self.bs, owner, operator)?;
-        Ok(())
+        self.transaction(|state, bs| {
+            let mut owner_map = state.get_owner_data_hamt(bs)?;
+
+            state.revoke_for_all(&mut owner_map, owner, operator)?;
+
+            state.owner_data = owner_map.flush().map_err(StateError::from)?;
```
anorth, line 329: This kind of thing clearly has no business logic so I think the old way was fine.

frcxx_nft/src/lib.rs
```diff
@@ -241,19 +341,40 @@ where
...
-            recipient,
-            token_ids,
+        self.transaction(|state, store| {
+            let mut token_array = state.get_token_data_amt(store)?;
+            let mut owner_map = state.get_owner_data_hamt(store)?;
+            NFTState::assert_owns_tokens(&token_array, owner_id, token_ids)?;
+
+            for &token_id in token_ids {
+                // update the token_data to reflect the new owner and clear approved operators
+                state.make_transfer(&mut token_array, &mut owner_map, token_id, recipient_id)?;
```
anorth, line 354: Pass the ownership test in a predicate here

anorth (commented):
Thanks, this is the right direction, but maybe went a fraction too far. To the extent that it's possible for the state level to maintain invariants (like consistency between the HAMTs), I _do_ think that belongs there. There might be some points where we have to sacrifice that for performance, but there's not yet one that stands out.

alexytsu (commented):
I'd suggest that for consistency no state functions should flush the hamt or amts

The improvement here is efficiency so that more complex use cases can batch multiple state updates (i.e. mint + transfer; transfer + burn) without wasted intermediary flushes to the blockstore.

It's slightly trickier to use (callers must remember to do the flush themselves) but for the general use case, this is already handled via the library level. Alternatively we could provide a transaction helper that lazily loads the Hamt/Amt as needed and automatically flushes them to the blockstore when the transaction is closed.

anorth:
That sounds good in principle, but maybe tricky in practise. It would be fine to add a `flush()` method to be called by the library or other user, and require it. But since the state object holds CIDs, how would you get the CIDs to hold between different mutations except by flushing? We'd need a separate object that holds instantiated h/amts (which have internal caches for unwritten blocks).

If we were to really push optimising for performance, this could make sense, but I'm not sure it's justified here yet. I think the fungible token library generally made a fair tradeoff here. We know we could add more complexity for more batch transaction performance there, but decided to leave that for later.

frcxx_nft/src/lib.rs
```diff
@@ -152,17 +179,49 @@ where
...
+        owner: &Address,
+        operator: &Address,
+        token_ids: &[TokenID],
+    ) -> Result<u64> {
+        let operator = self.msg.resolve_id(operator)?;
+        let owner = self.msg.resolve_or_init(owner)?;
+
+        let balance = self.transaction(|state, bs| {
+            let owner_map = state.get_owner_data_hamt(bs)?;
+            let account_operator = NFTState::is_account_operator(&owner_map, owner, operator)?;
```
alexytsu, line 208: checking this outside requires an extra load (same for transfer_from)

alternatively this logic can be written in the predicate which means it will be checked for each token. this is not so bad if the predicate has the internally pre-loaded hamt available.

alexytsu: https://github.com/helix-onchain/filecoin/issues/171 will track this here unless it's a blocker for merging this PR

frcxx_nft/src/lib.rs
```diff
@@ -125,15 +144,23 @@ where
         token_data: RawBytes,
     ) -> Result<ReceiverHook<MintIntermediate>> {
         let operator = self.msg.resolve_id(operator)?;
-        let initial_owner = self.msg.resolve_or_init(initial_owner)?;
-        Ok(self.state.mint_tokens(
-            &self.bs,
+        let initial_owner_id = self.msg.resolve_or_init(initial_owner)?;
+
+        let mint_intermediate = self.transaction(|state, bs| {
```
anorth, line 149: Transaction is unnecessary here and in other places now.

frcxx_nft/src/lib.rs
anorth: Not a blocker. But instead of getting the owner map out and then calling a function, make a new state method `is_account_operator(owner, operator)`

frcxx_nft/src/lib.rs
```diff
@@ -272,26 +349,48 @@ where
...
+                token_ids,
+                owner_id,
+                recipient_id,
+                &|token_data, token_id| {
+                    NFTState::assert_owns_token(token_data, token_id, owner_id)?;
+                    if !account_operator {
+                        NFTState::assert_token_level_approval(token_data, token_id, operator_id)
+                    } else {
+                        Ok(())
+                    }
```
anorth, line 379: There's a small bit of repeated logic here from transfer. If it's easy to factor out, let's do so, but no biggie if it's not

frcxx_nft/src/state.rs
```diff
@@ -234,47 +215,56 @@ impl NFTState {
 
     /// Approves an operator to transfer a set of specified tokens
     ///
-    /// Checks that the caller is the owner of the specified token. If any of the token_ids is not
-    /// valid (i.e. non-existent or not-owned by the caller), the entire batch approval is aborted.. If any of the token_ids is not
-    /// valid (i.e. non-existent or not-owned by the caller), the entire batch approval is aborted.
-    pub fn approve_for_tokens<BS: Blockstore>(
+    /// The caller should own the tokens or an account-level operator on the owner of the tokens.
```
anorth, line 218: I think all of these "the caller" comments are out of place here. This is a pure data object, the caller is some library code. Describe what it does, don't prescribe how to use it.

anorth (approved):

PR #167, opened 2022-12-06
SDK Shim

alexytsu:
Aims to replace the triple injection of Blockstore, Messaging, Actor functionality with a single interface for the NFT library. 

The FRC46 token library should follow suit later but must be supported for now - hence the duplication of code.


fvm_actor_utils/src/util.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    fn root_cid(&self) -> ActorResult<Cid> {
+        Ok(self.fvm_runtime.root()?)
+    }
+}
+
+impl Blockstore for FvmActor {
+    fn get(&self, cid: &Cid) -> anyhow::Result<Option<Vec<u8>>> {
+        // If this fails, the _CID_ is invalid. I.e., we have a bug.
+        fvm_sdk::ipld::get(cid)
```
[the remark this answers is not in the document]
alexytsu: I've made the runtimes themselves impl Blockstore as the ipld::get and ipld::set operations will be trickier to implement and will require a new FakeBlockstore impl rather than relying on MemoryBlockstore. 

Alternatively the Runtime trait could have a method that returns a reference to a Blockstore

fvm_actor_utils/src/runtime/test_runtime.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// A map of addresses that were instantiated in this runtime
+    pub addresses: RefCell<HashMap<Address, ActorID>>,
+    /// The next-to-allocate f0 address
+    pub next_actor_id: RefCell<ActorID>,
+
+    /// The last message sent via this runtime
+    pub last_message: RefCell<Option<RawBytes>>,
+    /// Flag to control message success
+    abort_next_send: RefCell<bool>,
```
anorth, line 27: `pub`?

frcxx_nft/src/lib.rs
```diff
@@ -110,8 +113,8 @@ where
 
     /// Return the number of NFTs held by a particular address
     pub fn balance_of(&self, address: &Address) -> Result<u64> {
-        let balance = match self.msg.resolve_id(address) {
-            Ok(owner) => self.state.get_balance(&self.bs, owner)?,
+        let balance = match self.helper.resolve_id(address) {
+            Ok(owner) => self.state.get_balance(&self.helper.runtime, owner)?,
```
anorth, line 117: The difference between sometimes calling a method on `helper` or sometimes extracting the `runtime` from it suggests something slightly off in this design. Either:
- abstract the runtime with the helper (make it private), forwarding methods, or
- turn the helper into a collection of free functions that take a runtime parameter, no struct; hold an `RT` object in the NFT struct.

The helper struct boils down to just currying in the runtime object to a bunch of helper functions, but I don't think it adds utility. I'd probably go for the second approach, but ok with either.

Then
```
        let balance = match util::resolve_id(self.runtime, address) {
            Ok(owner) => self.state.get_balance(&self.runtime, owner)?,
```

anorth (commented):
Just one design question now that we can see the whole thing. Otherwise looks great.

frcxx_nft/src/lib.rs
alexytsu: When you reviewed, I think the only places where `helper.runtime` was used was in methods that expected an `impl Blockstore` param. 

I've now made the `ActorHelper` itself implement the trait rather than having the runtime do it so the abstraction feels less leaky now

alexytsu:
@anorth ready for re-review but would prefer to land #176 into here before merging to main

anorth:
Ping me after merging #176

frc46_token/src/token/mod.rs
```diff
@@ -35,15 +37,13 @@ type Result<T> = std::result::Result<T, TokenError>;
...
+    S: Syscalls,
     BS: Blockstore,
-    MSG: Messaging,
 {
-    /// Injected blockstore. The blockstore must reference the same underlying storage under Clone
-    bs: BS,
-    /// Minimal interface to call methods on other actors (i.e. receiver hooks)
-    msg: MSG,
+    /// Runtime services to interact with the execution environment
+    pub runtime: ActorRuntime<S, BS>,
```
anorth, line 46: Can we avoid the `pub`? At least hide it behid a method

frc46_token/src/token/mod.rs
```diff
@@ -859,40 +857,46 @@ mod test {
...
         let cid = token.flush().unwrap();
 
         // the returned cid can be used to reference the same token state
-        let mut state = Token::<_, FakeMessenger>::load_state(&bs, &cid).unwrap();
-        let token2 =
-            Token::wrap(&bs, FakeMessenger::new(TOKEN_ACTOR.id().unwrap(), 6), 1, &mut state);
+        let helper = ActorRuntime {
+            blockstore: token.runtime.blockstore,
+            syscalls: FakeSyscalls::default(),
+        };
```
anorth, line 871: This is a bit unfortunate - the helper is consumed by `wrap`? This will probably strike in actor users outside tests, too.
- It may be appropriate to hold only a reference (w lifetime) to the helper in the token.
- Else it may be appropriate to add a clone() or similar method to the helper to create another one backed by the same things (requiring the syscalls/blockstore to be `+Clone`)

fvm_actor_utils/src/syscalls/mod.rs
```diff
@@ -0,0 +1,44 @@
+use cid::Cid;
+use fvm_ipld_encoding::ipld_block::IpldBlock;
+use fvm_shared::{
+    address::Address, econ::TokenAmount, error::ErrorNumber, receipt::Receipt, ActorID, MethodNum,
+};
+use thiserror::Error;
+
+mod fvm_syscalls;
+pub use fvm_syscalls::FvmSyscalls;
```
anorth, line 9: Will this prevent importing this in contexts where the SDK isn't available? Perhaps drop these convenience exports.

fvm_actor_utils/src/util.rs
```diff
@@ -0,0 +1,152 @@
...
+use fvm_shared::receipt::Receipt;
+use fvm_shared::MethodNum;
+use fvm_shared::METHOD_SEND;
+use fvm_shared::{address::Address, econ::TokenAmount, ActorID};
+use num_traits::Zero;
+use thiserror::Error;
+
+use crate::messaging::{Messaging, MessagingError, Result as MessagingResult};
+use crate::syscalls::FakeSyscalls;
+use crate::syscalls::FvmSyscalls;
```
anorth, line 14: Ditto I'm nervous this will prevent building in some contexts.

fvm_actor_utils/src/util.rs
```diff
@@ -0,0 +1,152 @@
...
+///
+/// It provides higher level utilities than raw syscalls for actors to use to interact with the
+/// IPLD layer and the FVM runtime (e.g. messaging other actors)
+#[derive(Clone, Debug)]
+pub struct ActorRuntime<S: Syscalls, BS: Blockstore> {
+    pub syscalls: S,
+    pub blockstore: BS,
+}
+
+impl<S: Syscalls, BS: Blockstore> ActorRuntime<S, BS> {
```
anorth, line 36: Please add a simple `new(syscalls, store)` method so the caller can choose deps

anorth (approved):
Looks good overall, mostly minor suggestions.

frc46_token/src/token/mod.rs
alexytsu: though the helper is consumed, it can be (re)accessed via `pub fn runtime(&self) -> ActorRuntime<S, BS>` 

this follows the same ownership and reference structure as TokenState

PR #168, opened 2022-12-06
FRC46 Token coverage

abright:
Fills some coverage gaps across the frc46_token library, to get it up over 90% coverage

### DIFF
--- a/frc46_token/src/token/state.rs
+++ b/frc46_token/src/token/state.rs
@@ -767,7 +767,7 @@ mod test {
     use fvm_shared::{bigint::Zero, ActorID};
 
     use super::TokenState;
-    use crate::token::state::{actor_id_key, StateError, StateInvariantError};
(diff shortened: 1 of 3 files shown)


frc46_token/src/token/state.rs
```diff
@@ -270,7 +270,14 @@ impl TokenState {
     /// This involves iterating through the entire HAMT
     pub fn count_balances<BS: Blockstore>(&self, bs: &BS) -> Result<usize> {
         let balance_map = self.get_balance_map(bs)?;
-        Ok(balance_map.for_each(|_, _| Ok(())).into_iter().count())
```
anorth, line 273: Was this a bug?

anorth (approved):
Solid!

frc46_token/src/token/state.rs
abright: Yes - `for_each` returns a `Result`, so it was converting that into an iterator and counting it, which always returned 1 regardless of the underlying tree size.

PR #169, opened 2022-12-06
Use Option<IpldBlock> for all message params

arajasek:
This PR:
- Updates FVM SDK and ipld_encoding
- Refactors to start using `Option<IpldBlock>` for message parameters

### DIFF
--- a/fvm_actor_utils/src/messaging.rs
+++ b/fvm_actor_utils/src/messaging.rs
@@ -2,8 +2,8 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 
 use frc42_dispatch::method_hash;
+use fvm_ipld_encoding::ipld_block::IpldBlock;
 use fvm_ipld_encoding::Error as IpldError;
-use fvm_ipld_encoding::RawBytes;
 use fvm_sdk::{actor, message, send, sys::ErrorNumber};
 use fvm_shared::error::ExitCode;
 use fvm_shared::receipt::Receipt;
@@ -61,7 +61,7 @@ pub trait Messaging {
         &self,
         to: &Address,
         method: MethodNum,
-        params: &RawBytes,
+        params: Option<IpldBlock>,
         value: &TokenAmount,
     ) -> Result<Receipt>;
 
@@ -129,10 +129,10 @@ impl Messaging for FvmMessenger {
         &self,
         to: &Address,
         method: MethodNum,
-        params: &RawBytes,
+        params: Option<IpldBlock>,
         value: &TokenAmount,
     ) -> Result<Receipt> {
-        Ok(send::send(to, method, params.clone(), value.clone())?)
+        Ok(send::send(to, method, params, value.clone())?)
     }
 
     fn resolve_id(&self, address: &Address) -> Result<ActorID> {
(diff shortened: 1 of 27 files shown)


fvm_actor_utils/src/receiver/mod.rs
```diff
@@ -119,10 +121,17 @@ impl<T: RecipientData> ReceiverHook<T> {
...
             RECEIVER_HOOK_METHOD_NUM,
-            &RawBytes::serialize(&params)?,
+            Some(IpldBlock::serialize(IPLD_RAW, &params_ser).map_err(|e| {
+                ReceiverHookError::IpldEncoding(fvm_ipld_encoding::Error {
+                    description: e.to_string(),
+                    protocol: fvm_ipld_encoding::CodecProtocol::Cbor,
+                })
+            })?),
             &TokenAmount::zero(),
         )?;
```
Stebalien, line 136: Why double-serialize?

Stebalien: Ah, I think this should be `Some(IpldBlock::serialize_cbor(params)?)`.

arajasek:
CI failures are unrelated, see: https://github.com/helix-onchain/filecoin/pull/170

anorth:
This library hasn't updated to fvm 3 yet. I'm interpreting this PR as sort of a prototype to check it will be ok to make this change, given the github dependencies etc.

arajasek:
@anorth This PR won't be updating to FVM 3 (that will have to happen, but later). The GH dependencies will be removed, of course, but I would appreciate a review of the direction. I don't want to land [the corresponding FVM PR](https://github.com/filecoin-project/ref-fvm/pull/1199) without some confidence that the downstream consumers will be able to update.

fvm_actor_utils/src/messaging.rs
```diff
@@ -152,7 +152,7 @@ impl Messaging for FvmMessenger {
 ///
 #[derive(Debug)]
 pub struct FakeMessenger {
-    pub last_message: RefCell<Option<RawBytes>>,
+    pub last_message: RefCell<Option<Option<IpldBlock>>>,
```
anorth, line 155: Do we really need `Option<Option<>>`? This at least needs documentation to explain it.

anorth (commented):
This looks fine directionally. I would approve when it's actually ready to land.

Can we get a helper method on IpldBlock, or just change `serialize_cbor` to return the `Some()` to reduce boilerplate here?

An alternative to evaluate for an FVM type to encapsulate `Option<IpldBlock>` so that most users don't have to know about this implementation detail, and avoid yet more boilterplate. These are both suggestions for the FVM rather than here.

alexytsu:
These changes look reasonable to me

the check-build CI error is a simple linting error and can be suppressed by cherry-picking bafec14d2cd0cc4bb1f4249ac31868954b2340f0 or rebasing of the latest `main` commit

the linking error when building actors is complaning about some missing `opencl` library. I can solve this locally by installing `ocl-icd-opencl-dev` on Ubuntu but I'm not sure what's changed that requires this to build the actors

anorth (approved):

PR #166, opened 2022-11-28
fix: allow method name start with  underscore

hunjixin:
method name can start with under score

https://github.com/filecoin-project/FIPs/blob/0044af63ebaa669a286750d4b2ca61ffceff52c3/FRCs/frc-0042.md?plain=1#L78

PR #172, opened 2022-12-12
FRC-0042 method name can start with underscore

geoff-vball:
Method names should be able to start with an underscore. This is detailed in the spec https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0042.md

### DIFF
--- a/frc42_dispatch/hasher/src/hash.rs
+++ b/frc42_dispatch/hasher/src/hash.rs
@@ -50,8 +50,8 @@ pub enum MethodNameErr {
 
 #[derive(Error, PartialEq, Eq, Debug)]
 pub enum IllegalNameErr {
-    #[error("method name doesn't start with capital letter")]
-    NotCapitalStart,
+    #[error("method name doesn't start with capital letter or _")]
+    IllegalStartingCharacter,
     #[error("method name contains letters outside [a-zA-Z0-9_]")]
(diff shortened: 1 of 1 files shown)


frc42_dispatch/hasher/src/hash.rs
```diff
@@ -108,10 +108,10 @@ fn check_method_name(method_name: &str) -> Result<(), MethodNameErr> {
         return Err(MethodNameErr::EmptyString);
     }
 
-    // Check starts with capital letter
+    // Check starts with capital letter or an underscore
     let first_letter = method_name.chars().next().unwrap(); // safe because we checked for empty string
-    if !first_letter.is_ascii_uppercase() {
-        return Err(IllegalNameErr::NotCapitalStart.into());
+    if !first_letter.is_ascii_uppercase() && first_letter != "_" {
```
anorth, line 113: Please apply DeMorgan's law to simplify the reading of operators
```
if !(first_letter.is_ascii_uppercase() || first_letter == '_')
```

anorth (commented):
This is a duplicate of #166. But has the advantage of being based closer to `main`. If you can refer to the changes there and update the tests accordingly, I'm happy to merge whichever. If we merge this one, we should also credit @hunjixin.

PR #176, opened 2022-12-20
update fungible token to use new shim

alexytsu:
Pulled these changes out separately from https://github.com/helix-onchain/filecoin/pull/167 as it's quite noisy and we might not want to head in this direction (i.e. change the existing fungible token APIs)

This PR de-dupes some code in the fvm_actor_utils package while maintaining lightweight alternatives (i.e. Messaging trait) to the full-blown `ActorHelper` utility.

However, i think it's a good indicator that the ActorHelper is a useful abstraction if it simplifies the existing code (which I feel it does)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,3 @@ members = [
     "testing/fil_token_integration/actors/frc46_test_actor",
     "testing/fil_token_integration/actors/frcxx_test_actor"
 ]
-
-[patch.crates-io]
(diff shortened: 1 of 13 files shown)


anorth (approved):
I agree.

"Helper" is not a great name, it's worth a few moments to find something better.
- "Context" might be barely better
- "Runtime" if it wasn't taken by a component of this helper – perhaps we could rename that to Syscalls or similar?
- "Environment"?
- "Frame"?

PR #177, opened 2022-12-23
feat: SDK::send returns ExitCode and Option<IpldBlock>

arajasek:
Integrates https://github.com/filecoin-project/ref-fvm/pull/1366


fvm_actor_utils/src/receiver/mod.rs
```diff
@@ -132,16 +147,16 @@ impl<T: RecipientData> ReceiverHook<T> {
             &TokenAmount::zero(),
         )?;
 
-        match receipt.exit_code {
+        match ret.0 {
```
Stebalien, line 150: Ok, I think we want to name the return struct. This is going to get confusing.

arajasek: Done :)

frc46_token/src/token/types.rs
```diff
@@ -118,7 +118,7 @@ pub struct MintReturn {
     pub balance: TokenAmount,
     /// The new total supply.
     pub supply: TokenAmount,
-    /// (Optional) data returned from receiver hook
+    /// (Optional) CBOR-encoded data returned from receiver hook
```
anorth, line 121: I don't think anything here forces this to be CBOR, and this comment is likely to rot. cc @Stebalien

anorth (commented):
Looks good after deps are clean

fvm_actor_utils/src/messaging.rs
```diff
@@ -51,6 +50,21 @@ impl From<&MessagingError> for ExitCode {
     }
 }
 
+/// Matches the SDK's Response type
+/// We duplicate the type here, because this trait must be implemented
```
Stebalien, line 54: trait?

frc46_token/src/token/types.rs
Stebalien: We should probably just switch this over to `Option<IpldBlock>` (or actually check the codec).

arajasek: https://github.com/helix-onchain/filecoin/issues/178

anorth (approved):

Stebalien (approved):

PR #180, opened 2023-01-06
Update the v10 actors testing bundle

arajasek:
Points to the latest commit in the builtin actors integ branch (https://github.com/filecoin-project/builtin-actors/pull/987), which in turn points back to the latest commit in `main` here. 

This reference should be updated once more when https://github.com/filecoin-project/builtin-actors/pull/987 lands.


alexytsu (approved):
TY

Issue #181, opened 2023-01-09
Run package releases off CI


alexytsu:
- The Github CI should be given permissions to publish to crates.io
- Merged results on `main` should be used
- Tags should be used to trigger the releases and mark the commit that was built

BigLep:
If I had my permissions, I would change this issue such that:

# Issue title
Releases are triggered by code review

# Issue description

## Done Criteria
The to-be-created release steps outlined in https://github.com/filecoin-project/builtin-actors/issues/1566 can be followed without requiring manual clicking in the GitHub UI.  Instead, releases should occur as a result of an approved PR.

## Why Important
1. Less time used by maintainers
2. Not bottleneck on certain folks to publish
3. Less error prone

## User/Customer
Maintainers

## Notes
1. This is more of a standard Rust repo and should be published to crates.io

PR #182, opened 2023-01-09
Release frc42_dispatch 3.0.0, frc_46 3.1.0, fvm_actor_utils 3.0.0

alexytsu:
To release will follow these steps: https://github.com/helix-onchain/filecoin/issues/181

In future, this should be automated so that anyone with sufficient permissions on the repository can publish releases easily.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,3 @@ members = [
     "frc46_factory_token",
     "frc46_factory_token/token_impl"
 ]
-
-[patch.crates-io]
-# tracking issue: https://github.com/bitvecto-rs/funty/issues/7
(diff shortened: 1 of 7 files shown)


alexytsu:
I don't have a good way to --dry-run the releases/updates here. For example, frc42_dispatch needs an updated frc42_hasher which is yet to be published. 

@abright any tips to check that all the underlying dependencies that need bumping have been touched?

anorth (approved):

abright:
> @abright any tips to check that all the underlying dependencies that need bumping have been touched?

Not sure if there's some nice automated way to do it, I check manually but that won't scale well if there's more than a couple dependencies involved. Bumping version numbers earlier (when changes are made) would probably let cargo catch it for us, though that puts us at the mercy of cargo's resolver so maybe it wouldn't trip up if we'd only made patch-level changes or something?

PR #185, opened 2023-01-17
WIP: AMT Experiments

alexytsu:
WIP: ~this is in conjunction with https://github.com/helix-onchain/ref-fvm/pull/1~
this is supported by changes from https://github.com/helix-onchain/ref-fvm/pull/2

- [x] allow AMT traversal to be interrupted once a certain number of blocks have been expanded
- [x] allow AMT traversal to be resumed from a particular path
- [ ] benchmark gas savings and tune the default pagination


alexytsu:
Updated API to specify a start_index and end_index rather than track number of nodes expanded.

frc53_nft/src/state.rs
```diff
@@ -557,20 +557,44 @@ impl NFTState {
...
+        let mut data = Vec::new();
+
+        token_data_array
+            .for_range_while(range_start, end_index, |id, token_data| {
+                token_ids.set(id);
+                data.push(token_data.clone());
+                Ok(true)
+            })
+            .unwrap();
+        Ok((token_ids, data, end_index))
```
anorth, line 596: See https://github.com/helix-onchain/ref-fvm/pull/2, but this needs to eventually return a sentinel (0? INT_MAX?) when there are no more items to list.

PR #190, opened 2023-01-31
FRC46 factory token unit test coverage, ActorRuntime improvements

abright:
This started out when I wanted to add some unit test coverage to `token_impl` for the factory token, and update it to use the `ActorRuntime` instead of raw syscalls. While doing this, I kept running into walls where the runtime was missing something I needed or (in the most annoying case) that I can't clone or share the thing around as I needed.

Summary of changes:
- ~~made `ActorRuntime` clone()-able (which required adding the trait bound everywhere)~~  replaced by making FRC46 `Token`take a reference to the `ActorRuntime` instead of owning it
- added `SharedMemoryBlockstore` wrapper for a ref-counted `MemoryBlockstore` we can clone and share with other runtime instances (for cloning when a temporary `Token` handle needs it)
- added caller ID to `ActorRuntime` including a method to set it in `FakeSyscalls` for unit testing
- added `set_root` to `ActorRuntime` so there's no need for raw syscalls when saving state and updating the root CID
- most of the `FactoryToken` implementation is now covered by unit tests (which pass param structs and use a fake runtime so they're closer to how the integration tests work, it's not a total repeat of everything we already cover in `frc46_token`)

### DIFF
--- a/frc46_factory_token/src/lib.rs
+++ b/frc46_factory_token/src/lib.rs
@@ -13,7 +13,8 @@ fn token_invoke(method_num: u64, params: u32) -> Result<u32, RuntimeError> {
     match_method!(method_num, {
         "Constructor" => {
             let params = deserialize_params(params);
-            construct_token(params)
+            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
+            construct_token(runtime, params)
         }
         "Mint" => {
             let root_cid = fvm_sdk::sself::root()?;
--- a/frc46_factory_token/src/lib.rs
+++ b/frc46_factory_token/src/lib.rs
@@ -1,4 +1,7 @@
 use frc42_dispatch::match_method;
+use fvm_actor_utils::{
+    blockstore::Blockstore, syscalls::fvm_syscalls::FvmSyscalls, util::ActorRuntime,
+};
 use fvm_sdk::NO_DATA_BLOCK_ID;
 use fvm_shared::error::ExitCode;
 use token_impl::{
@@ -15,13 +18,15 @@ fn token_invoke(method_num: u64, params: u32) -> Result<u32, RuntimeError> {
         "Mint" => {
             let root_cid = fvm_sdk::sself::root()?;
             let params: MintParams = deserialize_params(params);
-            let mut token_actor = FactoryToken::load(&root_cid)?;
+            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
+            let mut token_actor = FactoryToken::load(runtime, &root_cid)?;
             let res = token_actor.mint(params)?;
             return_ipld(&res)
         }
         "DisableMint" => {
             let root_cid = fvm_sdk::sself::root()?;
-            let mut token_actor = FactoryToken::load(&root_cid)?;
+            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
+            let mut token_actor = FactoryToken::load(runtime, &root_cid)?;
             // disable minting forever
             token_actor.disable_mint()?;
             // save state
@@ -33,7 +38,8 @@ fn token_invoke(method_num: u64, params: u32) -> Result<u32, RuntimeError> {
         _ => {
             let root_cid = fvm_sdk::sself::root()?;
 
-            let mut token_actor = FactoryToken::load(&root_cid)?;
+            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
+            let mut token_actor = FactoryToken::load(runtime, &root_cid)?;
 
             let res = frc46_invoke(method_num, params, &mut token_actor, |token| {
                 // `token` is passed through from the original token provided in the function call
--- a/frc46_factory_token/src/lib.rs
+++ b/frc46_factory_token/src/lib.rs
@@ -10,43 +10,39 @@ use token_impl::{
 };
 
 fn token_invoke(method_num: u64, params: u32) -> Result<u32, RuntimeError> {
+    let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
     match_method!(method_num, {
         "Constructor" => {
             let params = deserialize_params(params);
-            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
             construct_token(runtime, params)
         }
         "Mint" => {
-            let root_cid = fvm_sdk::sself::root()?;
+            let root_cid = runtime.root_cid()?;
             let params: MintParams = deserialize_params(params);
-            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
             let mut token_actor = FactoryToken::load(runtime, &root_cid)?;
             let res = token_actor.mint(params)?;
             return_ipld(&res)
         }
         "DisableMint" => {
-            let root_cid = fvm_sdk::sself::root()?;
-            let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
+            let root_cid = runtime.root_cid()?;
             let mut token_actor = FactoryToken::load(runtime, &root_cid)?;
             // disable minting forever
             token_actor.disable_mint()?;
             // save state
             let cid = token_actor.save()?;
-            fvm_sdk::sself::set_root(&cid)?;
+            token_actor.runtime().set_root(&cid)?;
             // no return
             Ok(NO_DATA_BLOCK_ID)
         }
         _ => {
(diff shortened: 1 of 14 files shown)


fvm_actor_utils/src/util.rs
```diff
@@ -27,25 +28,39 @@ type ActorResult<T> = std::result::Result<T, ActorError>;
 /// It provides higher level utilities than raw syscalls for actors to use to interact with the
 /// IPLD layer and the FVM runtime (e.g. messaging other actors)
 #[derive(Clone, Debug)]
-pub struct ActorRuntime<S: Syscalls, BS: Blockstore> {
+pub struct ActorRuntime<S: Syscalls + Clone, BS: Blockstore + Clone> {
```
abright, line 31: I don't like having the `Clone` trait bound being spammed everywhere. Going to try changing the `Token` handle in `frc46_token` to use a mutable borrow instead of taking ownership, as it does for the token state. They are typically created together and passed to the `Token` so it shouldn't have a large impact on the code and allows me to avoid the `clone` call that caused all this.

fvm_actor_utils/src/syscalls/fake_syscalls.rs
```diff
@@ -19,10 +19,13 @@ pub struct TestMessage {
 #[derive(Clone, Default, Debug)]
 pub struct FakeSyscalls {
     /// The root of the calling actor
-    pub root: Cid,
+    pub root: RefCell<Cid>,
     /// The f0 ID of the calling actor
```
anorth, line 23: Could you update this comment to say "receiving actor"?

fvm_actor_utils/src/util.rs
anorth: Yes, please do try that. This could be awkward for built-in actors too

frc46_factory_token/token_impl/src/lib.rs
```diff
@@ -99,24 +108,19 @@ pub struct ConstructorParams {
 }
 
 pub fn construct_token(params: ConstructorParams) -> Result<u32, RuntimeError> {
-    let runtime = ActorRuntime::<FvmSyscalls, Blockstore>::new_fvm_runtime();
+    let runtime = ActorRuntime::<FvmSyscalls, RuntimeBlockstore>::new_fvm_runtime();
```
anorth, line 111: It seems wrong that the FvmSyscalls is used concretely here, rather than just the trait. Shouldn't the runtime be passed in as a parameter?

frc46_factory_token/token_impl/src/lib.rs
```diff
@@ -252,54 +274,65 @@ pub struct MintParams {
...
         Ok(())
     }
 
     pub fn mint(&mut self, params: MintParams) -> Result<MintReturn, RuntimeError> {
         // check if the caller matches our authorise mint operator
         // no minter address means minting has been permanently disabled
-        let minter = self.minter.ok_or(RuntimeError::MintingDisabled)?;
-        let caller_id = sdk::message::caller();
+        let minter = self.state.minter.ok_or(RuntimeError::MintingDisabled)?;
+        let caller_id = self.runtime.caller(); // TODO: may need to add this to ActorRuntime
```
anorth, line 335: Remove TODO?

frc46_factory_token/token_impl/src/lib.rs
```diff
@@ -328,13 +361,13 @@ impl FactoryToken {
     pub fn disable_mint(&mut self) -> Result<(), RuntimeError> {
         // no minter means minting has already been permanently disabled
         // we return this if already disabled because it will make more sense than failing the address check below
-        let minter = self.minter.ok_or(RuntimeError::MintingDisabled)?;
-        let caller_id = sdk::message::caller();
+        let minter = self.state.minter.ok_or(RuntimeError::MintingDisabled)?;
+        let caller_id = self.runtime.caller(); // TODO: add this to ActorRuntime
```
anorth, line 365: Remove TODO?

anorth (commented):
I think we should be able to get rid of that Clone trait. Do the built-in actors offer any guidance here?

fvm_actor_utils/src/util.rs
alexytsu: While the Syscalls and Blockstore implementations should be made clone-able so that actors and unit tests etc. can share them between `Token` handles, I'm not sure that the ActorRuntime needs to require them to be the case in the trait bounds. 

I may be missing some of the use case, but as far as I can tell, nothing within the ActorRuntime actually uses the fact that the blockstore is cloneable.

alexytsu (commented):
The changes here all look like good improvements except as we've all noted, the introduction of `Clone` requirements. 

If that can be avoided then LGTM

frc46_factory_token/token_impl/src/lib.rs
abright: I'll pass in the `ActorRuntime` since it does look weird here. It's still only used in actor code though because it returns a block ID. It had to live here in the past to avoid import conflicts in the integration tests, but that's also fixed now.

fvm_actor_utils/src/util.rs
abright: Without the `Clone` trait bound, you can't actually call `clone()` on the `ActorRuntime` (it knows about cloning itself but as far as it understands, it's composed of components that can't clone). Changing the `Token` to use a reference instead seems to be working well so I've taken out all the clone stuff to use that instead.

fvm_actor_utils/src/syscalls/fake_syscalls.rs
abright: Done, along with the ID line below it to avoid confusion with the caller ID

.github/workflows/check-main.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+env:
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
```
abright, line 9: Just found out about this over the weekend and we're using a new enough nightly build that I thought it was worth a try. Looks like it might be shaving a couple mins off the CI build times compared to other PRs but I haven't done any proper timing or benchmarking of it yet.

anorth: Cool, but would you mind splitting this out to a separate PR we can evaluate independently?

frc46_factory_token/token_impl/src/lib.rs
```diff
@@ -135,21 +136,39 @@ pub struct FactoryToken {
...
+            Err(e) => Err(RuntimeError::Deserialization(e.to_string())),
+        }
+    }
+}
+
 /// Implementation of the token API in a FVM actor
 ///
 /// Here the Ipld parameter structs are marshalled and passed to the underlying library functions
-impl FRC46Token for FactoryToken {
+impl<SC: Syscalls + Clone, BS: Blockstore + Clone> FRC46Token for FactoryToken<SC, BS> {
```
anorth, line 160: Can we remove the `Clone` requirements here?

anorth (approved):
This looks solid. I'd appreciate review from @alexytsu , but LGTM.

frc46_factory_token/src/lib.rs
```diff
@@ -10,18 +13,21 @@ fn token_invoke(method_num: u64, params: u32) -> Result<u32, RuntimeError> {
     match_method!(method_num, {
         "Constructor" => {
             let params = deserialize_params(params);
-            construct_token(params)
```
alexytsu, line 13: nit: if the runtime is being instantiated in every match branch, perhaps move to before the `match_method!`

frc46_factory_token/token_impl/src/lib.rs
```diff
@@ -106,25 +110,22 @@ pub struct ConstructorParams {
...
+    params: ConstructorParams,
+) -> Result<u32, RuntimeError> {
     let minter = runtime.resolve_id(&params.minter)?;
     let token =
-        FactoryToken::new(&runtime, params.name, params.symbol, params.granularity, Some(minter));
+        FactoryToken::new(runtime, params.name, params.symbol, params.granularity, Some(minter));
 
     let cid = token.save()?;
-    fvm_sdk::sself::set_root(&cid)?;
+    fvm_sdk::sself::set_root(&cid).map_err(|_| NoStateError)?;
```
alexytsu, line 122: can this be swapped with a function from `ActorRuntime` since we have access to one here

fvm_actor_utils/src/util.rs
```diff
@@ -93,6 +108,11 @@ impl<S: Syscalls, BS: Blockstore> ActorRuntime<S, BS> {
         Ok(self.syscalls.root().map_err(|_err| NoStateError)?)
     }
 
+    // Set the root cid of the actor's state
```
alexytsu, line 111: nit: ///

alexytsu (approved):
LGTM

PR #194, opened 2023-02-10
Test: Set CI environment to use Cargo's new sparse index protocol

abright:
See https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html

Not sure if this will have much benefit for the way our CI builds, but it's worth testing to see if it does save a few minutes off the total runtime

### DIFF
--- a/.github/workflows/check-main.yml
+++ b/.github/workflows/check-main.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+env:
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
+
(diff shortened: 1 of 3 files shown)


abright:
Since this stuff only affects cargo downloading its index, it doesn't seem to have much impact on our build times. Most of the time is spent building the many dependencies we have and there's no easy solutions there. Some typical approaches are to run dedicated hardware or to go down the rabbit hole of building inside containers with pre-built and cached dependencies (and then automating that to update in step with `Cargo.lock`, etc)

PR #198, opened 2023-02-16
NFT Enumeration Features

alexytsu:
Requires review and release of AMT library https://github.com/filecoin-project/ref-fvm/pull/1673

- [x] Add tests at state-level to bump coverage score (the functionality is already covered at integration level)


frc53_nft/src/lib.rs
```diff
@@ -412,6 +415,56 @@ where
...
+    ) -> Result<ListTokensReturn> {
+        let owner_id = self.runtime.resolve_id(owner)?;
+        let (tokens, next_cursor) =
+            self.state.list_owned_tokens(&self.runtime, owner_id, cursor, Some(max))?;
+        Ok(ListTokensReturn { tokens, next_cursor })
+    }
+
+    /// Returns all the operators approved by an owner for a token
+    pub fn list_token_operators(&self, owner: &Address, token_id: TokenID) -> Result<Vec<ActorID>> {
+        let _owner_id = self.runtime.resolve_id(owner)?;
```
alexytsu, line 439: not sure if there is value in asserting that the given token is owned by a specific address, I suggest we modify the FIP to not require this parameter.

else I'll add an additional check at the state-level functionality

anorth: Great catch, thank you.

frc53_nft/src/lib.rs
```diff
@@ -412,6 +415,56 @@ where
...
+        Ok(self.state.list_token_operators(&self.runtime, token_id)?)
+    }
+
+    /// Enumerates tokens for which an account is an operator for an owner
+    pub fn list_operator_tokens(
+        &self,
+        owner: &Address,
```
anorth, line 446: Does the same apply here, that we should just iterate tokens for which an account is an operator, without specifying an owner. Otherwise how would the caller figure out which owners are worth passing here anyway?

frc53_nft/src/state.rs
```diff
@@ -553,18 +585,134 @@ impl NFTState {
...
+    /// List all the minted tokens
+    pub fn list_tokens<BS: Blockstore>(
+        &self,
+        bs: &BS,
+        range_start: Option<AmtCursor>,
+        limit: Option<u64>,
+    ) -> Result<(TokenSet, Option<AmtCursor>)> {
+        let token_data_array = match &range_start {
+            Some(cursor) => self.get_token_amt_at_cursor(bs, cursor)?,
+            None => self.get_token_data_amt(bs)?,
```
anorth, line 610: Nit: factor out this repeated pattern, perhaps by passing the optional cursor to get_toke_data_amt

frc53_nft/src/state.rs
```diff
@@ -553,18 +585,134 @@ impl NFTState {
...
+        let next_cursor = next_key.map(|key| AmtCursor::new(self.token_data, key));
+        Ok((token_ids, next_cursor))
+    }
+
+    /// List all the token operators for a given token_id
+    pub fn list_token_operators<BS: Blockstore>(
+        &self,
+        bs: &BS,
+        token_id: TokenID,
+    ) -> Result<Vec<ActorID>> {
```
anorth, line 659: You convinced me to make this a paginated result in the FRC.

frc53_nft/src/state.rs
```diff
@@ -553,18 +585,134 @@ impl NFTState {
...
+
+    /// Enumerates tokens for which an account is an operator for a given owner
+    pub fn list_operator_tokens<BS: Blockstore>(
+        &self,
+        bs: &BS,
+        owner: ActorID,
+        operator: ActorID,
+        cursor: Option<AmtCursor>,
+        max: Option<u64>,
+    ) -> Result<ListOperatorTokensReturn> {
```
anorth, line 676: Ditto paginated result

frc53_nft/src/state.rs
```diff
@@ -553,18 +585,134 @@ impl NFTState {
...
+        tokens.iter().try_for_each(|token_id| -> Result<()> {
+            let token_data = token_array.get(token_id)?.ok_or_else(|| {
+                StateError::InvariantFailed(format!(
+                    "token {token_id} owned by {owner} not found in token amt",
+                ))
+            })?;
+
+            if token_data.operators.get(operator) {
+                operatable_tokens.set(token_id);
+            }
```
anorth, line 692: If we remove the `owner` parameter here I expect this will simplify down to look a lot like list_owned_tokens.

frc53_nft/src/state.rs
```diff
@@ -553,18 +585,134 @@ impl NFTState {
...
+
+        Ok(ListOperatorTokensReturn { tokens: operatable_tokens, next_cursor })
+    }
+
+    /// List all the token operators for a given account
+    pub fn list_account_operators<BS: Blockstore>(
+        &self,
+        bs: &BS,
+        actor_id: ActorID,
+    ) -> Result<Vec<ActorID>> {
```
anorth, line 705: Paginated result here too

frc53_nft/src/types.rs
```diff
@@ -159,3 +162,48 @@ pub struct RevokeParams {
 pub struct RevokeForAllParams {
     pub operator: Address,
 }
+
+#[derive(Serialize_tuple, Deserialize_tuple, Clone, Debug)]
+pub struct ListTokensParams {
+    pub cursor: Option<AmtCursor>,
```
anorth, line 168: For the purposes of an API decoupled from implementation, I think we want all the cursor parameters to be `[u8]`, from/into which you can then de/serialise an AmtCursor explicitly. This should probably remain `[u8]` at the library level, but be translated for the state API.

frc53_nft/src/lib.rs
alexytsu: Yeh, makes sense to remove `owner` param here

frc53_nft/src/types.rs
alexytsu: I disagree here given that this particular cursor format _is_ inherently tied to this specific reference implementation/library. The library method could take `[u8]` as a parameter but it's going to break anyway if anything other than a serialised `AmtCursor` is passed in so the typesafety at library level does add some value. The API as specified by the FRC dictates actor-actor boundaries, against which `ListTokensParams` is compliant when serialised (if my understanding of the serialisation logic is correct).


If your motivation is more about communicating that the `cursor` can be implementation-specific, then I'd rather add some documentation/examples demonstrating alternate or augmented cursor mechanics.

anorth: Firstly context: I'm interpreting these types as the ones that should be used by actors implementing tokens to define their binary API. Otherwise I would not expect derived serialization. The built-in datacap actor uses those types already defined here.

This is not compliant with the FRC (which specified IPLD-CBOR tuple encoding). The FRC specifies something of the form
```
[
 h'1234abcd',
 10000
]
```
This will produce instead
```
[
 [
  42(h'5678efgh'),
  100
 ],
 10000
]
```

Where the bytes in `h'1234abcd'` would deserialize to `[42(h'5678efgh'),100]`, and `42(...)` is a CID. (Use cbor.me to inspect those snippets.)

I understand that it will be possible to pass ListTokensParams that deserialize correctly but contain an invalid token. That's fine – there are other ways to get an invalid token too. The library can deserialize the opaque token and check its structure.

alexytsu: Made the changes https://github.com/helix-onchain/filecoin/pull/198/commits/85e812eaed6b83dd1dac5d2cd2108946d0fd38f6

anorth (commented):
Thanks, looks good. If the H/AMT enumeration features are in FVM now, are we just waiting for a release before this is done done?

alexytsu:
Yeh - just need AMT to be released. HAMT also needs a release but iirc is not actually used here yet.

alexytsu:
@anorth I've switched the deps off of git references to the published crates now.

I'll publish a v1 of the NFT crates when this is landed and go review the FIP to make any final updates there.

anorth (approved):

PR #204, opened 2023-02-28
add documentation to set_balance

alexytsu:


frc46_token/src/token/state.rs
```diff
@@ -229,6 +229,8 @@ impl TokenState {
     }
 
     /// Set the balance of the account returning the old balance
+    ///
+    /// Does NOT update total supply, the caller must do this as a separate step if necessary
```
anorth, line 233: 👍 , and I think we can say a bit more.

"Consistent with change_balance_by, this method does not change the total supply. Business logic to reconcile the total supply with changes in balances is the responsibility of the caller"

anorth (approved):

PR #212, opened 2023-06-28
chore: update fvm/ipld crates

Stebalien:
I'm also moving all the version specifications to the workspace to ensure we always use the same versions and to make future updates easier.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,14 @@ members = [
     "testing/fil_token_integration/actors/frc46_test_actor",
     "testing/fil_token_integration/actors/frc53_test_actor"
 ]
+
+[workspace.dependencies]
+cid = { version = "0.10.1", default-features = false, features = ["serde-codec"] }
+fvm_sdk = "~3.3.0"
+fvm_shared = "~3.4.0"
+fvm_integration_tests = "~3.1.0"
+fvm_ipld_encoding = "0.4.0"
+fvm_ipld_blockstore = "0.2.0"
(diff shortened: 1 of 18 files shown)


Stebalien:
See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table for some context.

Stebalien:
So, this is going to be painful. Basically:

- Our wasm builder version doesn't work with modern rust compiler versions.
- We require modern rust compilers (wasmtime requires them, specifically).
- Newer wasm builder versions compile the "wasm-opt" program and make compile times _very_ **VERY** slow.

We've solved this in ref-fvm by manually building wasm the same way we do in the builtin actors. I have a partial solution to this but..., well, I'll file a PR to demonstrate the issue.

alexytsu:
> I'm also moving all the version specifications to the workspace to ensure we always use the same versions and to make future updates easier.

TIL, thanks.

> We've solved this in ref-fvm by manually building wasm the same way we do in the builtin actors. I have a partial solution to this but..., well, I'll file a PR to demonstrate the issue.

I was following https://github.com/filecoin-project/ref-fvm/pull/1718 and https://github.com/filecoin-project/ref-fvm/pull/1698 to update to stable rust here but didn't quite get all the way yet (on branch `alex/stable-toolchain`)

Stebalien:
Ah! Awesome. This is the branch I currently have: https://github.com/helix-onchain/filecoin/tree/steb/remove-wasm-builder.

PR #215, opened 2023-06-29
replace the `no_sdk` feature with `use_sdk`

lemmih:
Compile features work better when they are additive.

### DIFF
--- a/frc42_dispatch/hasher/src/hash.rs
+++ b/frc42_dispatch/hasher/src/hash.rs
@@ -1,5 +1,3 @@
-#[cfg(not(feature = "no_sdk"))]
-use fvm_sdk::crypto;
 use thiserror::Error;
 
 /// Minimal interface for a hashing function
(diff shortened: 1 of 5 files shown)


lemmih:
Before this PR, the `no_sdk` feature was independent from the dependencies and compiling without default features would fail:
```
$ cargo build --no-default-features
   Compiling frc42_hasher v1.5.0 (/home/lemmih/coding/filecoin/frc42_dispatch/hasher)
error[E0432]: unresolved import `fvm_sdk`
 --> frc42_dispatch/hasher/src/hash.rs:2:5
  |
2 | use fvm_sdk::crypto;
  |     ^^^^^^^ use of undeclared crate or module `fvm_sdk`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `frc42_hasher` due to previous error
```

With this PR, all feature combinations work as intended. Compiling without default features disables the SDK code, and compiling with default features (or explicitly enabling `use_sdk`) will pull in the right dependencies.

frc42_dispatch/hasher/src/hash.rs
```diff
@@ -16,15 +14,15 @@ pub trait Hasher {
...
+    #[cfg(feature = "use_sdk")]
     fn hash(&self, bytes: &[u8]) -> Vec<u8> {
         use fvm_shared::crypto::hash::SupportedHashes;
-        crypto::hash_owned(SupportedHashes::Blake2b512, bytes)
+        fvm_sdk::crypto::hash_owned(SupportedHashes::Blake2b512, bytes)
     }
 
     // stub version that avoids fvm_sdk syscalls (eg: for proc macro, or built-in actor use)
-    #[cfg(feature = "no_sdk")]
+    #[cfg(not(feature = "use_sdk"))]
```
alexytsu, line 25: I understand this is not directly related to your change @lemmih but while you are here I think we can drop this `unimplemented` code block entirely if we compile Blake2bSyscall and the impl Hasher for Blake2bSyscall blocks conditionally on `use_sdk`.

alexytsu (approved):
Thanks for the improvement

frc42_dispatch/hasher/src/hash.rs
lemmih: Done. Let me know what you think.

PR #216, opened 2023-07-05
Bump version numbers for release

alexytsu:
Preps new version numbers to land once we have switched to stable rust

TODO

- [x] land stable rust toolchain and build wasm actors with new method

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,11 @@ fvm_ipld_hamt = "0.7.0"
 fvm_ipld_amt = { version = "0.6.0", features = ["go-interop"] }
 fvm_ipld_bitfield = "0.5.4"
 
+[patch.crates-io]
+fvm_actor_utils = { git = "https://github.com/helix-onchain/filecoin", branch = "alex/update-fvm-ipld" }
+frc42_dispatch = { git = "https://github.com/helix-onchain/filecoin", branch = "alex/update-fvm-ipld" }
+frc46_token = { git = "https://github.com/helix-onchain/filecoin", branch = "alex/update-fvm-ipld" }
+
 [profile.wasm]
 inherits = "release"
 panic = "abort"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,17 @@ members = [
     "testing/test_actors/actors/frc46_factory_token/token_impl",
 ]
 
+[workspace.dependencies]
+cid = { version = "0.10.1", default-features = false, features = ["serde-codec"] }
+fvm_sdk = "~3.3.0"
+fvm_shared = "~3.4.0"
+fvm_integration_tests = "~3.1.0"
+fvm_ipld_encoding = "0.4.0"
+fvm_ipld_blockstore = "0.2.0"
+fvm_ipld_hamt = "0.7.0"
+fvm_ipld_amt = { version = "0.6.0", features = ["go-interop"] }
+fvm_ipld_bitfield = "0.5.4"
+
 [profile.wasm]
 inherits = "release"
 panic = "abort"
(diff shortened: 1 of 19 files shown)


alexytsu:
@Stebalien I'm a little confused as to the versioning of crates such as fvm_shared, fvm, fvm_sdk. 

Within a given project, is it recommended/necessary to have only one minor version of any of these particular crates? I.e. if a project pulls in dependencies that specify both fvm_shared@3.2 and fvm_shared@3.3 is this ok or are they considered incompatible? 

In that case, in this repo we generally bump major versions when upgrading these dependencies to signal to our consumers that these changes are incompatible and need to be updated in line with the fvm crates. However, would bumping minor versions in a PR here be sufficient?

Stebalien:
> In that case, in this repo we generally bump major versions when upgrading these dependencies to signal to our consumers that these changes are incompatible and need to be updated in line with the fvm crates. However, would bumping minor versions in a PR here be sufficient?

Yeah.... See https://github.com/filecoin-project/ref-fvm/issues/1724.

Basically, we've been versioning the FVM 1, 2, 3, etc. with for major incompatible _network_ upgrades. Which means API breaking changes get minor version bumps. And yeah, this is incorrect.

My goal is to turn all of this into `fvm1*` and `fvm2*` crates using major versions for API-breaking changes. But that work keeps getting delayed behind more pressing stuff.

alexytsu:
I'm slightly stuck here as we need builtin-actors https://github.com/filecoin-project/builtin-actors/pull/1330 to be updated to the same versions of FVM in order to build integration tests here. (Though technically we should be able to release the packages off this branch without passing the integration tests).

Cargo.toml
```diff
@@ -16,6 +16,22 @@ members = [
...
+fvm_ipld_encoding = "0.4.0"
+fvm_ipld_blockstore = "0.2.0"
+fvm_ipld_hamt = "0.7.0"
+fvm_ipld_amt = { version = "0.6.0", features = ["go-interop"] }
+fvm_ipld_bitfield = "0.5.4"
+
+[patch.crates-io]
+fvm_actor_utils = { git = "https://github.com/helix-onchain/filecoin", branch = "alex/update-fvm-ipld" }
+frc42_dispatch = { git = "https://github.com/helix-onchain/filecoin", branch = "alex/update-fvm-ipld" }
+frc46_token = { git = "https://github.com/helix-onchain/filecoin", branch = "alex/update-fvm-ipld" }
```
alexytsu, line 33: we need builtin-actors https://github.com/filecoin-project/builtin-actors/pull/1330 to be updated to the same versions of FVM in order to build integration tests here. (Though technically we should be able to release the packages off this branch without passing the integration tests). 

That builtin-actors commit patches back to this branch in order to bump its use of helix-libraries alongside underlying FVM versions. The issue now is that helix-filecoin depends on bactors (as a git patch) which depends back on helix-filecoin (as a git patch). However, when building _from_ helix-filecoin, the bactors -> helix-filecoin dep is trying to pull in the helix crates from crates.io. Hence here we have to patch back to ourselves.

After the helix crates are released to crates.io we should then
- remove this
- remove the patch for helix libraries in https://github.com/filecoin-project/builtin-actors/pull/1330
- land https://github.com/filecoin-project/builtin-actors/pull/1330
- patch our integration tests here to the `master` commit of bactors

alexytsu:
@Stebalien could you ptal

Stebalien (approved):
LGTM. One thing we should consider would be fixing the builtin actor's build system to apply patches when building actors. Substrate's wasm-builder does that but we were going for a "clean build". Bug given that the build isn't reproducible anyways.... it likely doesn't matter.

Stebalien:
> After the helix crates are released to crates.io we should then

Note: I'm not sure how `cargo` will react if you try to publish with the patch in-place. You may have to release with broken CI.

PR #217, opened 2023-07-05
Stable Toolchain

alexytsu:
- move integration tests in the src tree
- migrate frc46_test_actor to new build
- migrate integration test actors to the new build script

TODO in followup: https://github.com/helix-onchain/filecoin/issues/218


rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "stable"
```
Stebalien, line 2: So, the downside of specifying stable is that CI will break every time rust releases a new version and adds additional warnings. The upside is that you instantly get notified of new warnings/issues, it can just be a little annoying when you're trying to merge a PR and suddenly CI decides to break for unrelated reasons).

Stebalien (approved):

PR #225, opened 2023-09-06
chore: release crates for FVM update

Stebalien:
I've also marked some test crates as "don't publish" to make it easier to see what needs to be released.

This patch releases:

- frc42_dispatch 4.0.0
- frc42_hasher 2.0.0
- frc46_token 8.0.0
- frc53_nft 2.0.0
- fvm_actor_utils 8.0.0
- frc42_macros 2.0.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,8 +37,8 @@ num-traits = { version = "0.2.15" }
 anyhow = { version = "1.0.56" }
 
 # internal deps of published packages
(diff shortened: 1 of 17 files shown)


anorth:
I think we should bump major versions. If frc42_dispatch is matching FVM versions, it's only by coincidence.

Stebalien:
Ok. Major versions released.

alexytsu (approved):

PR #235, opened 2024-07-23
ci: automate creation of GitHub releases and tags

galargh:
This automates the creation of GitHub releases and associated GitHub tags based on the changes to the version in the Cargo.toml files across the repo. The tags will be prefixed with the directory in which the Cargo.toml file is. For example, changing the version in `dispatch_examples/greeter/Cargo.toml` will create `dispatch_examples/greeter/${VERSION}` tag.

The automation uses the same principles as the one you might know from Go projects where we control the current version via version.json file (https://github.com/ipdxco/unified-github-workflows?tab=readme-ov-file#versioning). This is the exact same workflow just updated to workflow with Cargo.toml.

Note, that the automation doesn't take care of cargo publishing currently.

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -2,19 +2,8 @@ name: Release Checker
 
 on:
   pull_request_target:
-    paths: [
-      "dispatch_examples/greeter/Cargo.toml",
-      "frc42_dispatch/Cargo.toml",
-      "frc42_dispatch/hasher/Cargo.toml",
-      "dispatch_examples/greeter/Cargo.toml",
-      "frc42_dispatch/Cargo.toml",
-      "frc42_dispatch/hasher/Cargo.toml",
-      "frc42_dispatch/macros/Cargo.toml",
-      "frc46_token/Cargo.toml",
-      "frc53_nft/Cargo.toml",
-      "fvm_actor_utils/Cargo.toml",
-      "fvm_dispatch_tools/Cargo.toml"
-    ]
+    paths:
+      - "**/Cargo.toml"
     types: [ opened, synchronize, reopened, labeled, unlabeled ]
   workflow_dispatch:
 
@@ -32,7 +21,6 @@ jobs:
     with:
       sources: |
         [
-          "dispatch_examples/greeter/Cargo.toml",
           "frc42_dispatch/Cargo.toml",
           "frc42_dispatch/hasher/Cargo.toml",
           "frc42_dispatch/macros/Cargo.toml",
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -6,22 +6,14 @@ on:
       "dispatch_examples/greeter/Cargo.toml",
       "frc42_dispatch/Cargo.toml",
       "frc42_dispatch/hasher/Cargo.toml",
+      "dispatch_examples/greeter/Cargo.toml",
+      "frc42_dispatch/Cargo.toml",
+      "frc42_dispatch/hasher/Cargo.toml",
       "frc42_dispatch/macros/Cargo.toml",
-      "frc42_dispatch/macros/example/Cargo.toml",
       "frc46_token/Cargo.toml",
       "frc53_nft/Cargo.toml",
       "fvm_actor_utils/Cargo.toml",
-      "fvm_dispatch_tools/Cargo.toml",
-      "testing/integration/Cargo.toml",
-      "testing/test_actors/Cargo.toml",
-      "testing/test_actors/actors/basic_nft_actor/Cargo.toml",
-      "testing/test_actors/actors/basic_receiving_actor/Cargo.toml",
-      "testing/test_actors/actors/basic_token_actor/Cargo.toml",
-      "testing/test_actors/actors/basic_transfer_actor/Cargo.toml",
-      "testing/test_actors/actors/frc46_factory_actor/Cargo.toml",
-      "testing/test_actors/actors/frc46_test_actor/Cargo.toml",
-      "testing/test_actors/actors/frc53_test_actor/Cargo.toml",
-      "testing/test_actors/actors/frc46_factory_token/token_impl/Cargo.toml"
+      "fvm_dispatch_tools/Cargo.toml"
     ]
     types: [ opened, synchronize, reopened, labeled, unlabeled ]
   workflow_dispatch:
@@ -44,19 +36,9 @@ jobs:
           "frc42_dispatch/Cargo.toml",
           "frc42_dispatch/hasher/Cargo.toml",
           "frc42_dispatch/macros/Cargo.toml",
-          "frc42_dispatch/macros/example/Cargo.toml",
           "frc46_token/Cargo.toml",
           "frc53_nft/Cargo.toml",
           "fvm_actor_utils/Cargo.toml",
-          "fvm_dispatch_tools/Cargo.toml",
-          "testing/integration/Cargo.toml",
-          "testing/test_actors/Cargo.toml",
-          "testing/test_actors/actors/basic_nft_actor/Cargo.toml",
-          "testing/test_actors/actors/basic_receiving_actor/Cargo.toml",
-          "testing/test_actors/actors/basic_token_actor/Cargo.toml",
-          "testing/test_actors/actors/basic_transfer_actor/Cargo.toml",
-          "testing/test_actors/actors/frc46_factory_actor/Cargo.toml",
-          "testing/test_actors/actors/frc46_test_actor/Cargo.toml",
-          "testing/test_actors/actors/frc53_test_actor/Cargo.toml",
-          "testing/test_actors/actors/frc46_factory_token/token_impl/Cargo.toml"
+          "fvm_dispatch_tools/Cargo.toml"
         ]
+      separator: "@"
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -1,0 +1,62 @@
+name: Release Checker
+
+on:
+  pull_request_target:
+    paths: [
+      "dispatch_examples/greeter/Cargo.toml",
+      "frc42_dispatch/Cargo.toml",
+      "frc42_dispatch/hasher/Cargo.toml",
+      "frc42_dispatch/macros/Cargo.toml",
+      "frc42_dispatch/macros/example/Cargo.toml",
+      "frc46_token/Cargo.toml",
+      "frc53_nft/Cargo.toml",
+      "fvm_actor_utils/Cargo.toml",
+      "fvm_dispatch_tools/Cargo.toml",
+      "testing/integration/Cargo.toml",
+      "testing/test_actors/Cargo.toml",
+      "testing/test_actors/actors/basic_nft_actor/Cargo.toml",
+      "testing/test_actors/actors/basic_receiving_actor/Cargo.toml",
+      "testing/test_actors/actors/basic_token_actor/Cargo.toml",
+      "testing/test_actors/actors/basic_transfer_actor/Cargo.toml",
+      "testing/test_actors/actors/frc46_factory_actor/Cargo.toml",
+      "testing/test_actors/actors/frc46_test_actor/Cargo.toml",
+      "testing/test_actors/actors/frc53_test_actor/Cargo.toml",
+      "testing/test_actors/actors/frc46_factory_token/token_impl/Cargo.toml"
+    ]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  release-check:
+    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
+    with:
+      sources: |
+        [
+          "dispatch_examples/greeter/Cargo.toml",
+          "frc42_dispatch/Cargo.toml",
+          "frc42_dispatch/hasher/Cargo.toml",
+          "frc42_dispatch/macros/Cargo.toml",
+          "frc42_dispatch/macros/example/Cargo.toml",
+          "frc46_token/Cargo.toml",
+          "frc53_nft/Cargo.toml",
+          "fvm_actor_utils/Cargo.toml",
+          "fvm_dispatch_tools/Cargo.toml",
+          "testing/integration/Cargo.toml",
+          "testing/test_actors/Cargo.toml",
+          "testing/test_actors/actors/basic_nft_actor/Cargo.toml",
+          "testing/test_actors/actors/basic_receiving_actor/Cargo.toml",
+          "testing/test_actors/actors/basic_token_actor/Cargo.toml",
+          "testing/test_actors/actors/basic_transfer_actor/Cargo.toml",
+          "testing/test_actors/actors/frc46_factory_actor/Cargo.toml",
+          "testing/test_actors/actors/frc46_test_actor/Cargo.toml",
+          "testing/test_actors/actors/frc53_test_actor/Cargo.toml",
+          "testing/test_actors/actors/frc46_factory_token/token_impl/Cargo.toml"
+        ]
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,18 +4,23 @@ This document describes the process for releasing a new version of the `actors-u
 
 ## Current State
 
-1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
-    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
-    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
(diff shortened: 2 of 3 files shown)


BigLep (commented):
Conceptually seems right but I'll leave to those who have done hands-on releasing here to approve like @rjan90 or @rvagg or @Stebalien.

Also, for the record this addresses someo f https://github.com/filecoin-project/actors-utils/issues/181 , but it isn't handling artifact creating/publishing (as called out in the PR description).

rvagg:
`frc42_dispatch/hasher` and `frc42_dispatch/macros` would be new packages that haven't previously been tagged here if those Cargo.toml files are touched.

Same for `frc42_dispatch/macros/example`, `dispatch_examples/greeter`, `testing/integration`, `testing/test_actors`, and all of the `testing/test_actors/actors/*` packages.

Is this reasonable and OK? I have to defer to @Stebalien || @anorth on what we expect to be released from here.

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,62 @@
...
+      "testing/integration/Cargo.toml",
+      "testing/test_actors/Cargo.toml",
+      "testing/test_actors/actors/basic_nft_actor/Cargo.toml",
+      "testing/test_actors/actors/basic_receiving_actor/Cargo.toml",
+      "testing/test_actors/actors/basic_token_actor/Cargo.toml",
+      "testing/test_actors/actors/basic_transfer_actor/Cargo.toml",
+      "testing/test_actors/actors/frc46_factory_actor/Cargo.toml",
+      "testing/test_actors/actors/frc46_test_actor/Cargo.toml",
+      "testing/test_actors/actors/frc53_test_actor/Cargo.toml",
+      "testing/test_actors/actors/frc46_factory_token/token_impl/Cargo.toml"
```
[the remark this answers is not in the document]
galargh: Unfortunately, that isn't that straighforward.

galargh:
@rvagg thanks for the note. I took the list of packages we might want handled from https://github.com/filecoin-project/actors-utils/blob/c2e3fcdb8a81961c03d53a53310117b616798206/Cargo.toml#L3 If some should be omitted, please let me know.

Also, as per the description, this doesn't handle cargo publishing at the moment. The most straightforward way to handle that would be to follow a similar flow to what we proposed in - https://github.com/filecoin-project/builtin-actors/pull/1571#issuecomment-2248299241 There, the post-GitHub Release publishing step is to upload assets to the release; here, it could be to perform cargo publishing.

.github/workflows/release-check.yml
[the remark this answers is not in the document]
BigLep: Ack - bummer.  I guess this in part because GitHub Actions doesn't support YML anchors?
Maybe for now just add a comment that any changes here should also be made in releaser.yml (and vice versa)

BigLep:
Looking at this more, I think we need @Stebalien input.  I'll point him at this.

On the surface, I would assume we should just tag what has previously been tagged.  Looking at https://github.com/filecoin-project/actors-utils/tags, that appears to be:
frc42_dispatch
frc42_hasher
frc42_macros
frc46_token
frc53_nft
fvm_actor_utils
fvm_dispatch_tools

I also see this repo hasn't traditionally created GitHub releases: https://github.com/filecoin-project/actors-utils/releases .  I don't know if that's intentional or not.

anorth:
> On the surface, I would assume we should just tag what has previously been tagged

Yes, I think that would be the right place to start. Tagging/releasing/publishing crates only used for testing doesn't seem helpful.

Stebalien:
Yeah, two bits of feedback:

1. I'd much prefer to tag based on package names instead of directories.
2. Let's not tag/publish crates marked as `publish = false` in the Cargo.toml file.

Note: the `cargo metadata` command (specifically, `cargo metadata --no-deps` is particularly useful (no parsing toml).

@rvagg 

> frc42_dispatch/hasher and frc42_dispatch/macros would be new packages that haven't previously been tagged here if those Cargo.toml files are touched.

We have been publishing those as `frc42_hasher` and `frc42_macros`.

galargh:
I now limited the list of packages that are to be released with the automation and also updated the separtor to "@" so the packages will be tagged as `crate_name@crate_version`.

Here's the current list of config files that will be looked at:
```
      "dispatch_examples/greeter/Cargo.toml",
      "frc42_dispatch/Cargo.toml",
      "frc42_dispatch/hasher/Cargo.toml",
      "dispatch_examples/greeter/Cargo.toml",
      "frc42_dispatch/Cargo.toml",
      "frc42_dispatch/hasher/Cargo.toml",
      "frc42_dispatch/macros/Cargo.toml",
      "frc46_token/Cargo.toml",
      "frc53_nft/Cargo.toml",
      "fvm_actor_utils/Cargo.toml",
      "fvm_dispatch_tools/Cargo.toml"
```

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,44 @@
...
+      "frc42_dispatch/hasher/Cargo.toml",
+      "dispatch_examples/greeter/Cargo.toml",
+      "frc42_dispatch/Cargo.toml",
+      "frc42_dispatch/hasher/Cargo.toml",
+      "frc42_dispatch/macros/Cargo.toml",
+      "frc46_token/Cargo.toml",
+      "frc53_nft/Cargo.toml",
+      "fvm_actor_utils/Cargo.toml",
+      "fvm_dispatch_tools/Cargo.toml"
+    ]
```
Stebalien, line 17: Any reason we mention more paths here than below (`sources`)?

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,44 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  release-check:
+    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
+    with:
+      sources: |
```
Stebalien, line 33: Can we use yaml references/anchors to deduplicate this list?

Stebalien (commented):
This works but it requires duplicating information already present in the `Cargo.toml` files, which is unfortunate. Is there any way to extract this information from those files?

If that's difficult, having a single list of "released" crates would likely be good enough, especially if we can code-gen it.

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,44 @@
...
+        [
+          "dispatch_examples/greeter/Cargo.toml",
```
Stebalien, line 35: remove me

.github/workflows/release-check.yml
Stebalien: Alternatively, at least for the `paths:` part, we can probably just glob on `**/Cargo.toml`. We'll get some false positives, but that'll be fine.

BigLep:
2024-08-01 conversation between IPDX and FilOz: @galargh will incorporate comments week of 2024-08-05 so we can get this merged.

.github/workflows/release-check.yml
galargh: Good spot, a faulty copy-and-paste. I'll get that fixed.

.github/workflows/release-check.yml
galargh: Unfortunately, no advanced yaml features for GitHub Actions. We can definitely do `**/Cargo.toml`. I like this idea - will add to this branch.

galargh:
I applied the requested changes in https://github.com/filecoin-project/actors-utils/pull/235/commits/58afb97df022e620685bdec4c27bb9de06f7e128. I'll now proceed to add a similar commit to the PR in `ref-fvm`. Let me know if there's anything else you think this needs.

BigLep:
@Stebalien : I believe this is ready for your review again.

@galargh : can you add a README section about the release process like we have in some of the repos so it's clear to someone about the set of steps to take?  Something like:

```
1. Open a PR adjusting `version` in any of the Cargo.toml files.  This in turn causes X to occur.
2. When the PR is is merged, then X happens.
3. Afterwards you need to manually need to upload assets until https://github.com/filecoin-project/actors-utils/issues/181 is addressed
```
(I don't know if the above is accurate.)

Or drop me a couple of notes and I can happily do it.  I'm actually confused what how much of the release process is manual vs. automated, especially since this repo doesn't have GitHub Releases.  I assume I'm just missing things given being tired and having confusion compared to the other recent Rust repos we've been touching.

Stebalien (approved):
As far as this goes, it seems reasonable but won't actually be _useful_ until we can:

1. Publish on release.
2. Perform a dry-run before releasing.

~Just double checking: the tags are now `crate_name@version`?~ I see it is.

galargh:
> As far as this goes, it seems reasonable but won't actually be _useful_ until we can:
> 
> 1. Publish on release.
> 2. Perform a dry-run before releasing.
> 
> ~Just double checking: the tags are now `crate_name@version`?~ I see it is.

Yes, agreed 💯 

I described the process proposed here and future improvements in https://github.com/filecoin-project/actors-utils/pull/235/commits/e4daf658551110cdc250d8d9869034d55642c410.

I think we should decide whether the partial automation is useful for this repo or we should wait until we can also fully automate the `cargo publish` steps.

RELEASE.md
```diff
@@ -0,0 +1,28 @@
+# Release Process
+
+This document describes the process for releasing a new version of the `actors-utils` project.
+
+## Current State
+
+1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
```
BigLep, line 8:
```suggestion
    1. Extract the version from the modified `Cargo.toml` files, and process each crate in the workspace **independently**.
```

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
+    3. Create a draft GitHub release with the version as the tag.
+    4. Comment on the pull request with a link to the draft release.
+2. **[MANUAL]** Run `cargo publish --dry-run` for each crate that is proposed to be released in the reverse dependency order.
+3. On pull request merge, a [Releaser](.github/workflows/release.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
```
BigLep, line 15: Similar to other places asked, but in the happy/you-are-following-the-process case, the tag will already be created in step 1.2 right?

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+## Current State
+
+1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
+    3. Create a draft GitHub release with the version as the tag.
+    4. Comment on the pull request with a link to the draft release.
+2. **[MANUAL]** Run `cargo publish --dry-run` for each crate that is proposed to be released in the reverse dependency order.
+3. On pull request merge, a [Releaser](.github/workflows/release.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
```
BigLep, line 14:
```suggestion
    1. Extract the version from the modified `Cargo.toml` files, and process each crate in the workspace **independently**.
```

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+4. **[MANUAL]** Run `cargo publish` for each crate that has been released in the reverse dependency order.
+
+#### Known Limitations
+
+1. `cargo publish --dry-run` has to be run manually.
+2. `cargo publish` has to be run manually.
+
+#### Possible Improvements
+
+1. Run `cargo publish --dry-run` in the [reverse dependency order](#crate-dependency-graph) automatically. Use a local registry to simulate the dependencies that are not yet published.
```
BigLep, line 27: Is `crate-dependency-graph` defined anywhere?

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+This document describes the process for releasing a new version of the `actors-utils` project.
+
+## Current State
+
+1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
+    3. Create a draft GitHub release with the version as the tag.
+    4. Comment on the pull request with a link to the draft release.
+2. **[MANUAL]** Run `cargo publish --dry-run` for each crate that is proposed to be released in the reverse dependency order.
```
BigLep, line 12: How does someone find the reverse dependency order?

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+3. On pull request merge, a [Releaser](.github/workflows/release.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
+    3. Check if a draft GitHub release with the version as the tag exists.
+    4. If the draft release exists, publish it. Otherwise, create a new release with the version as the tag.
+4. **[MANUAL]** Run `cargo publish` for each crate that has been released in the reverse dependency order.
```
BigLep, line 18:
```suggestion
4. **[MANUAL]** Run `cargo publish` for each crate that has been released in the reverse dependency order.
   - You may need to ask one of the owners listed in https://crates.io/crates/fvm_actor_utils so you can publish.
```

@Stebalien : this is me guessing that there is permission access requirements.  Assuming there is, can you add more of the Lotus streamline pod who would be helping do releases here?

BigLep (commented):
Thanks for documenting.  This seems better than what we had before in that have more docs and a few less manual steps. I'll let @Stebalien make the call whether this is worth it or not though.

RELEASE.md
```diff
@@ -0,0 +1,28 @@
+# Release Process
+
+This document describes the process for releasing a new version of the `actors-utils` project.
+
+## Current State
+
+1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
```
BigLep, line 7:
```suggestion
1. Create a pull request which updates the [`version` in one or more `Cargo.toml` files](https://github.com/search?q=repo%3Afilecoin-project%2Factors-utils+path%3ACargo.toml+%2F%5Eversion+%3D%2F&type=code).
   - Title the PR `build: release Ɑ for Ƅ` where:
      - `Ɑ` is what is being released like "all crates"
      - `Ƅ` is the cause for the release like "fvm vX.Y.Z" or "nvXX"
2. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
```

I wanted to make clear what files are to be edited and to give guidance on how to name the PR.

RELEASE.md
BigLep: PR to update who is part of fvm-crate-owners: https://github.com/filecoin-project/github-mgmt/pull/58

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
```
rvagg, line 9:
```suggestion
    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists, and continue only if it does not.
```
just to be consistent with the above suggestion.

RELEASE.md
rvagg: And how does this get done for the crate? we have a workspace here so do we need to `-p` to do this or do we `cd` into the directory to run this? I'm not clear on how publishing on workspaces works so this is a bit of a mystery to me.

RELEASE.md
rvagg:
```suggestion
    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists, and continue only if it does not.
```

galargh: The tag gets created on GitHub Release publish.

BigLep:
Closing this PR for identical reasons to https://github.com/filecoin-project/ref-fvm/pull/2027#issuecomment-2290194343 .  Bummer :(

For next network upgrade, nv24, if the situation hasn't improved, we'll have someone like @rjan90 or @BigLep watch @Stebalien and capture the manual steps that are covered so we at least document the current manual release process better. We'll also make sure that others have access to publish by adjusting https://crates.io/crates/fvm_actor_utils and getting https://github.com/filecoin-project/github-mgmt/pull/58 landed.

BigLep:
(Turns out I don't permission to close this.)

galargh:
Same as in https://github.com/filecoin-project/ref-fvm/pull/2027#issuecomment-2306731873, thank you all for your time and knowledge - truly appreciated 🙇

PR #236, opened 2024-08-05
fix: address some clippy nits and warnings

Stebalien:
This skips the docs issues, I'm addressing that in a separate PR.

### DIFF
--- a/frc53_nft/src/lib.rs
+++ b/frc53_nft/src/lib.rs
@@ -587,13 +587,7 @@ mod test {
         {
             // mint no tokens
             let mut hook = nft
-                .mint(
-                    &ALICE,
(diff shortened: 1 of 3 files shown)


Stebalien:
Clippy is still unhappy because we also need #237. But I didn't want to do both of these in a single PR. CI should pass in https://github.com/filecoin-project/actors-utils/tree/steb/make-clippy-happy (combines both PRs).

Stebalien:
Ah, actually, we're not using clippy. We're just using `cargo check`. So https://github.com/filecoin-project/actors-utils/pull/237 fixes all the cargo check errors.

We'll have to add clippy, but we can do that later.

rvagg (approved):

PR #237, opened 2024-08-05
doc: fix indentation, punctuation, and links

Stebalien:
Clippy was complaining about list indentation and paragraph separation, so I fixed that and a few other issues (links to types, etc.).

I may have gone a little ham but... I couldn't stop once I started.

### DIFF
--- a/frc53_nft/src/types.rs
+++ b/frc53_nft/src/types.rs
@@ -7,87 +7,94 @@ use fvm_ipld_encoding::RawBytes;
 use fvm_shared::address::Address;
 use fvm_shared::ActorID;
 
+#[cfg(doc)]
+use super::state::Cursor;
+
 pub type TokenID = u64;
 
 /// Multiple token IDs are represented as a BitField encoded with RLE+ the index of each set bit
-/// corresponds to a TokenID
+/// corresponds to a TokenID.
 pub type TokenSet = BitField;
 
 /// Multiple actor IDs are represented as a BitField encoded with RLE+ the index of each set bit
-/// corresponds to a ActorID
+/// corresponds to a ActorID.
 pub type ActorIDSet = BitField;
 
-/// A trait to be implemented by FRC-0053 compliant actors
+/// A trait to be implemented by FRC-0053 compliant actors.
 pub trait FRC53NFT {
-    /// A descriptive name for the collection of NFTs in this actor
+    /// A descriptive name for the collection of NFTs in this actor.
     fn name(&self) -> String;
 
-    /// An abbreviated name for NFTs in this contract
+    /// An abbreviated name for NFTs in this contract.
     fn symbol(&self) -> String;
 
-    /// Gets a link to associated metadata for a given NFT
+    /// Gets a link to associated metadata for a given NFT.
     fn metadata(&self, params: TokenID) -> Cid;
 
-    /// Gets the total number of NFTs in this actor
+    /// Gets the total number of NFTs in this actor.
     fn total_supply(&self) -> u64;
 
     /// Burns a given NFT, removing it from the total supply and preventing new NFTs from being
-    /// minted with the same ID
-    fn burn(&self, params: TokenID);
+    /// minted with the same ID.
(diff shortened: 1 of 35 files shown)


Stebalien:
You can test this locally with `cargo check --all --all-targets` (and `cargo doc`). That should make sure there are no dangling links, at the very least.

Other than that, you'll have to read it to make sure I didn't make any glaring mistakes.

I found most of the missing periods with:

```bash
rg -t rust -U '^ *///.*[^.:]\n *(///\s*[A-Z\n]|[^/ ])'
```

frc53_nft/src/types.rs
```diff
@@ -7,87 +7,94 @@ use fvm_ipld_encoding::RawBytes;
 use fvm_shared::address::Address;
 use fvm_shared::ActorID;
 
+#[cfg(doc)]
+use super::state::Cursor;
```
rvagg, line 11: hah, so we have to import them for "doc" if they're not already imported here in order for the doc links to work?

frc53_nft/src/types.rs
```diff
@@ -97,13 +104,14 @@ impl RecipientData for MintIntermediate {
     }
 }
 
-/// Intermediate data used by transfer_return to construct the return data
+/// Intermediate data used by [`NFT::transfer_return`][`super::NFT::transfer_return`] to construct
```
rvagg, line 107: what's going on with both of these links? is this a typo that they're both included?

frc53_nft/src/util.rs
```diff
@@ -28,14 +28,16 @@ impl OperatorSet for BitField {
...
+/// Maintains set-like invariants in-memory by maintaining sorted order of the underlying array.
+///
+/// Insertion and deletion are O(n) operations but we expect operator lists to be a relatively small
+/// size.
+// TODO: benchmark this against some other options such as
+// - BTreeSet in memory, Vec serialized
+// - BTreeSet in memory and serialization
+// - HashSets...
+// - Hamt<ActorID, ()>
+// - Amt<ActorID>
```
rvagg, line 40: this is going to work OK? `//` after the `///`? should you just stuck the TODO bit inside the `impl` maybe?

rvagg (approved):
are the `.`'s actually necessary to satisfy clippy or are we just being pedantic?

frc53_nft/src/types.rs
Stebalien: Yep: https://doc.rust-lang.org/rustdoc/advanced-features.html#cfgdoc-documenting-platform-specific-or-feature-specific-information

Well, I could also link to it with the full path, but then all the links become rather long.

frc53_nft/src/types.rs
Stebalien: It's markdown formatting (reference style link). This is the explicit linking I was avoiding in https://github.com/filecoin-project/actors-utils/pull/237#discussion_r1704844288.

Rustdoc has some interesting handling of links: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html#linking-to-items-by-name

(I guess the backticks around the reference aren't necessary, but they work)

frc53_nft/src/util.rs
Stebalien: I can move it inside, I just didn't want it in the doc comment.

Stebalien:
> are the .'s actually necessary to satisfy clippy or are we just being pedantic?

No, I was just being pedantic. I like to ocasionally browse rustdocs for our projects and this:

```rust
/// This is some text
/// This is some other text
```

Will get formatted as:

> This is some text This is some other text

The only thing I _needed_ here was to fix:

```rust
/// - some very long bulletpoint item.
/// continuation...
```

Which needs to be:

```rust
/// - some very long bulletpoint item.
///   continuation...
```

frc53_nft/src/util.rs
Stebalien: I moved it above to make it clear.

PR #241, opened 2024-11-21
chore: remove default features from multihash-codetable

Stebalien:
This doesn't (yet) appear to make a difference but, once we release & update the builtin actors again, we may be able to drop a few dependencies from our tree? Regardless, it can't hurt.


rvagg (approved):
sgtm

PR #242, opened 2024-11-22
chore: switch back to builtin-actors master

Stebalien:
Now that the breaking cid/multihash changes have been merged, we can switch back.


rvagg (approved):
yar

PR #256, opened 2025-06-05
update frc46 version

jennijuju:


BigLep:
@rjan90 : are you going to handle merging this? 

(Also, happy to discuss separately, but should we get more people on actors-maintainers to help unblock these kinds of things quicker?)

rjan90:
> @rjan90 : are you going to handle merging this?
> (Also, happy to discuss separately, but should we get more people on actors-maintainers to help unblock these kinds of things quicker?)

I can take care of merging this, but I’ll need to refresh my memory on the release process, since I haven’t handled it before. As far as I understand, it mainly involves merging the version update and publishing the crates—which I believe I now have permissions for. However, I’m a bit unsure about the circular dependency that seems to involve actor-utils<>fvm<>actors.

BigLep:
Thanks for handling @rjan90 .  Per verbal, lets capture the steps / learnings into a release process document so we can follow/expand for the future.

rjan90:
Crates and tags published:
- https://crates.io/crates/frc46_token/14.1.0
- https://github.com/filecoin-project/actors-utils/tree/frc46_token%40v14.1.0

PR #258, opened 2025-06-11
docs: document release process

rjan90:
Attempt at documenting the release process - still some open TODO's here - that I do not have the answer for yet. But opening for initial review

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,67 @@ A configurable actor that can be used as a factory to create instances of
 tokens, based on [frc46_token](./frc46_token/README.md) and implemented
 [here](./testing/test_actors/actors/frc46_factory_token/)
 
+## Release Process
+
+This section documents the release process for actor-utils packages.
+
+### Published Packages
+
+The following packages are published to [crates.io](https://crates.io):
+
+- **`fvm_actor_utils`** - Core utilities for FVM native actors
+- **`frc42_dispatch`** - Method dispatch macros and utilities  
+- **`frc42_hasher`** - FRC-0042 method hashing utilities
+- **`frc42_macros`** - FRC-0042 procedural macros
+- **`frc46_token`** - Fungible token reference implementation
+- **`frc53_nft`** - Non-fungible token reference implementation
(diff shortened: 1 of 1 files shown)


BigLep:
> Can we consider folding this into the builtin-actors themselves?

Good point.  I really don't know rust tooling or the scope of the different crates to answer that.  I think it's a good question to ask to make sure we're not missing out on a simplification option.

Even if this moves into builtin-actors, the fact that we have subcrates means we still have some sequence of events "dance" that needs to be followed right?  Newbie question, but does the process simplify when everything is in the same repo?  I assume it's nice because there is one PR showing all the version updates, but then the act of pushing crates needs to be done in a certain order right?

Oh, or maybe you're saying to:
1. just pull everything into `builtin-actors` and 
2. 2nstead of publishing subcrates just publish new versions of `builtin-actors` when needed and any consumers of `actors-utils` subcrates today would update their dependencies to depend on `builtin-actors` as a whole.  
?

README.md
```diff
@@ -55,6 +55,67 @@ A configurable actor that can be used as a factory to create instances of
...
+   ```bash
+   # Tag the release
+   git tag <package-name>@<version>
+   git push origin <package-name>@<version>
+   ```
+
+#### Coordination Points
+
+- **FVM Releases**: TODO - Document coordination process with FVM releases
+- **Built-in Actors / Network Upgrade**: TODO - Document alignment with built-in actors and Network Upgrades
```
BigLep, line 117:
```suggestion
#### Coordination Points

Note: before documenting the coordination points more, we should first see about simplifying this repo by potentially removing unused functionality and moving some of it into actors-utils itself.  This is discussed in [issue #203](https://github.com/filecoin-project/actors-utils/issues/203).

- **FVM Releases**: TODO - Document coordination process with FVM releases
- **Built-in Actors / Network Upgrade**: TODO - Document alignment with built-in actors and Network Upgrades
```

BigLep (approved):
I added one note pointing to https://github.com/filecoin-project/actors-utils/issues/203

I think better to ship something here than to wait for the ideal repo cleanup to come about.

